PR #21, opened 2015-04-29
Update Library Importing method

irJERAD:
Updated the suggested method for importing the library (tested on multiple platforms and from many ways to be sure)
The "Add Library" function is no longer available the newest IDE. 
Additionally downloading the 6MB file from your site does not work (using multiple methods listed on your repo as well as the "official way" from the Arduino site) - results in Arduino IDE not being able to open and producing no warning. Downloading the 6.4MB file from the Repo works as far as I have tested using the method edited above.
Really appreciate your project and hope as I explore it more I might be able to provide more value for such a cool library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,11 @@ Download Mozzi from the top of this page and unzip it. It will probably have a
 name like "sensorium-Mozzi-2bee818". Rename the unzipped folder "Mozzi". Then,
 following the instructions from the [Arduino libraries guide](http://arduino.cc/en/Guide/Libraries):  
 
-In the Arduino IDE, navigate to __Sketch > Import Library__. At the top of the drop
-down list, select the option to __Add Library__. Return to the __Sketch > Import Library__ menu. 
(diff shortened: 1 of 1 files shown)


irJERAD:
Tried the new method from multiple vectors and believe I accounted for as many user types as possible.
Thanks for sharing your work, hope to be of more assistance in the future to this project as I get more time to play with the library

sensorium:
Thanks Jerad,
there is another big problem using Mozzi in Arduino IDE > 1.0.5, to do with
a difference in compilation which results in slower running code, causing
glitches in most exampke sketches.  Possibly a flag changed.... a real
problem unless there's a way to change the Arduino compilation, which I 'm
not aware of....

I think the problem installing in newer IDEs is because the
library.properties is incorrect and the folder layout is old-style - but
you've obviously looked at it more than I have, so thanks, I'll try to
release an updated version when I can find time.

Tim

sensorium:
I think this is resolved so I'll close it.

PR #22, opened 2015-05-09
Pr/21

sensorium:


sensorium:
Thanks for spotting the problem, the issue is fixed by just having a link to the Arduino library instructions.

PR #35, opened 2017-12-07
Add STM32 port

tfry-git:
Ok, maybe it's a bit too early to actually "request you to pull" this port, but I figured I'd better let you know about it early, and I think a pull request is a good way to organize the discussion.

As the subject states, this adds support for the STM32duino platform (maple based). It is not complete. HIFI_MODE and STEREO_HACK are missing completely, so far, many things need testing, and for the time being, we even need two small patches to the STM32 platform files. But, essentially - it's working, and I'm quite happy with the results so far! Details at the top of the README.md in my fork, and some initial discussion at http://www.stm32duino.com/viewtopic.php?f=2&t=2931#p37992 and http://www.stm32duino.com/viewtopic.php?f=15&t=2936 .

Enjoy!

### DIFF
--- a/AudioConfigSTM32.h
+++ b/AudioConfigSTM32.h
@@ -6,7 +6,7 @@
 #endif
 
 // Audio output pin. If you want to change this, make sure to also set AUDIO_PWM_TIMER to whichever timer is responsible for your PWM pin, and set the other timers to non-conflicting values
-#define AUDIO_CHANNEL_1_PIN PB6
+#define AUDIO_CHANNEL_1_PIN PB8
 #define AUDIO_PWM_TIMER 4
 // The timers used for running the control and audio update loops
 #define CONTROL_UPDATE_TIMER 3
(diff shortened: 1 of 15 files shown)


sensorium:
Nice work, tfry.  Should I wait a while before merging, or are you ready now?

tfry-git:
Glad you like it, and thanks the wonderful Mozzi library itself!

I'd say wait a few more days. Things I still intend to complete / wait for:
- HIFI mode (although some STM32 board have a dedicated DAC, the cheap ones don't, and I'm targeting the generic case. They are plenty fast, so the "default" mode gives 10bits at 70kHz modulation frequency on a single pin, but a few more bits, and an even faster modulation frequency will be nice). 
- It may be worth waiting until the two patches that I need in the STM32duino core are integrated, officially
- One (small) thing that I would like feedback on is how to handle mozziAnalogRead(): STM32 provides 12bit resolution, here (0-4095). Should I scale that down to the AVR range (0-1023), by default, for compatibility with existing sketches?

Finally, please note that I did some superficial cross-testing on a Pro Mini, but I do not have a Teensy to test with. Not sure what (how much) I broke, there...

tfry-git:
Hi again!

Ok, I think it's now "safe" to proceed: There is still one compilation patch pending on the STM32duino side, but that one is trivial, while the other that I was not so sure about (addition of "atomic.h"-header) is in. I also implemented HIFI mode and did some cleanups. I think things are in a pretty presentable state, now.

There are a few things that you will want to watch out for, before / while you merge, though:

1. I have prefixed my copy of README.md with a note about the purpose of my fork. You'll want to remove that, of course.
2. I've added a new section near the bottom of the README, with some details (and some caveats) for the new port.
3. While at it, I've also copied the Teensy 3.x notes to that section. You may want to second check that I used the right version of those, in case some of that has changed.
4. I _have_ tested compiling my fork for Arduino Pro Mini, and my simple Synth sketch continues to run as expected, there. So I'm relatively confident that I did not break anything for AVR.
5. I have _not_ testing compiling for Teensy 3.x, and even if I had, I don't have a Teensy to test with. I certainly tried to be careful, but I cannot guarantee that I did not break anything, there. (If I did it will probably be easy to fix - maybe I misplaced an "#if IS_TEENSY3()" or two, or things like that.)

Regards
Thomas

PR #37, opened 2018-01-09
Explicitly specify 16 bit ints for audio output bits and analog readings

tfry-git:
While playing with adding a larger audio output buffer (I thought it might help at higher audio rates, but did not actually reduce my troubles), I noted that on the 32bit platforms, the audio output buffer uses 32bit storage, as "int" is 32bit wide, here. This causes a waste of 512 bytes of RAM right there (twice as much for stereo) while adding virtually no benefit (to be entirely precise, using 32bit storage will be minimally faster, due to being already aligned, but the overhead of unaligned data is negligible compared to overhead of the timer interrupt; however, 512 bytes of RAM is not negligible in most cases).

This patch changes that to int16_t. On AVR boards, this does not make any difference. **On Teensy 3.x and STM32, existing sketches will break** as the signature of "int updateAudio()" is not the same as "int16_t updateAudio()", here.

The patch also addresses a similar issue in mozziAnalogRead(), although the amount of RAM saved is much smaller, here (but also sketches will remain source compatible in that respect).

### DIFF
--- a/MozziGuts.cpp
+++ b/MozziGuts.cpp
@@ -73,9 +73,9 @@ PWM frequency tests
 
 //-----------------------------------------------------------------------------------------------------------------
 // ring buffer for audio output
-CircularBuffer <unsigned int> output_buffer; // fixed size 256
+CircularBuffer <uint16_t> output_buffer; // fixed size 256
 #if (STEREO_HACK == true)
-CircularBuffer <unsigned int> output_buffer2; // fixed size 256
+CircularBuffer <uint16_t> output_buffer2; // fixed size 256
 #endif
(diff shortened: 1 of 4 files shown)


tfry-git:
Hmm. Thinking about it some more, I guess it will be better to leave this decision to the user by means of an appropriate define. I'll try to come up with something, shortly.

adrianfreed:
There are some unsigned longs in the code base too which should be reviewed with an eye to the larger platforms. we don't need phase precision better than 24bits in most musical applications so uint32_t is good enough.

tfry-git:
Obsoleted by #243

PR #38, opened 2018-01-11
Trigger updateControl from within audioHook(), rather than from timer.

tfry-git:
Ok, this one is definitely just a discussion starter, rather than a real pull request. It works for me, beautifully, on both STM32 and a Pro Mini, but it's clearly a non-trivial change. Also, if you were to give the go-ahead to this one, it would mean that _a lot_ of code can be removed (everything about Timer 0, essentially).

But to start at the beginning: The starting point to this idea was that I was struggling with some IO latency issues while handling MIDI file playback (see http://www.stm32duino.com/viewtopic.php?f=19&t=3041&sid=dc7ca666b2b92884cde768b1f4de1728). In that context I began thinking about what parts of the processing I could move from updateControl() to the main loop. But then, I started wondering, why isn't updateControl() part of the main loop in the first place?

Reasons that I could think up were:
1. updateControl() needs to be called at a defined rate. --> True, but at least for audio updates, that rate would actually be defined in terms of audio-ticks, not absolute time. This patch moves from being as exact on the time as possible to being perfectly exact on the audio-tick.
2. updateControl() should be able to run in parallel with updateAudio(). --> Yes, that would be nice, but it isn't actually the case. updateAudio() may be interrupted by updateControl() at any point, _but_ not the other way around. Thus, whenever updateControl() is running, updateAudio() is already blocked. Nothing changes, here.
3. Audio output needs to be updated while updateControl() is running. --> Yes, and that works because the audio output is updated from an interrupt (on timer 1), _not_ because updateControl() is run from an interrupt. Nothing changes, here.
4. Users might be doing fancy stuff like keeping some IO buffers filled from the main loop, while updateControl() reads those. --> No, see 2. Main loop and updateControl() do not actually run in parallel, at all. Besides, now users will have one additional timer free to _really_ implement something like this.
5. ???

Other aspects:
- For now the code assumes that AUDIO_RATE is evenly divisible by control rate, but it would not be hard to handle a fractional ratio.
- Besides removing timer 0, the patch should also allow to remove most (all?) uses of ATOMIC_BLOCK. That and the fact that interrupt calling overhead is removed, should provide a minor boost to performance as a nice plus.
- I am not quite sure, why, but this change, alone, also helps a lot with the latency issues I was having with MIDI file playback (see above).

What do you think? If you agree on the principle, then I could put some work into removing obsolete code, and updating the docs.

### DIFF
--- a/ADSR.h
+++ b/ADSR.h
@@ -17,7 +17,6 @@
 #else
  #include "WProgram.h"
 #endif
-//#include <util/atomic.h>
 #include "Line.h"
 #include "mozzi_fixmath.h"
 
--- a/AudioConfigSTM32.h
+++ b/AudioConfigSTM32.h
@@ -8,9 +8,8 @@
 // Audio output pin. If you want to change this, make sure to also set AUDIO_PWM_TIMER to whichever timer is responsible for your PWM pin, and set the other timers to non-conflicting values
 #define AUDIO_CHANNEL_1_PIN PB8
 #define AUDIO_PWM_TIMER 4
-// The timers used for running the control and audio update loops
-#define CONTROL_UPDATE_TIMER 3
-#define AUDIO_UPDATE_TIMER 2
+// The timer used for running the audio update loop
+#define AUDIO_UPDATE_TIMER 3
 
 #if (AUDIO_MODE == HIFI)
 // Second out pin for HIFI mode. This must be on the same timer as AUDIO_CHANNEL_1_PIN!
--- a/EventDelay.h
+++ b/EventDelay.h
@@ -13,7 +13,7 @@
 #define EVENTDELAY_H_
 
 
-/** A non-blocking replacement for Arduino's delay() function (which is disabled by Mozzi). 
+/** A non-blocking replacement for Arduino's delay() function. 
 EventDelay can be set() to a number of milliseconds, then after calling start(), ready() will return true when the time is up.  
 Alternatively, start(milliseconds) will call set() and start() together.
 */
--- a/Line.h
+++ b/Line.h
@@ -17,7 +17,6 @@
 #else
  #include "WProgram.h"
 #endif
-#include <util/atomic.h>
 
 /** For linear changes with a minimum of calculation at each step. For instance,
 you can use Line to make an oscillator glide from one frequency to another,
@@ -58,10 +57,7 @@ public:
 	inline
 	T next()
 	{
-		ATOMIC_BLOCK(ATOMIC_RESTORESTATE)
-		{
-			current_value += step_size;
-		}
+		current_value += step_size;
 		//Serial.println(current_value);
 		return current_value;
 	}
@@ -76,10 +72,7 @@ public:
 	inline
 	void set(T value)
 	{
-		ATOMIC_BLOCK(ATOMIC_RESTORESTATE)
-		{
-			current_value=value;
-		}
+		current_value=value;
 	}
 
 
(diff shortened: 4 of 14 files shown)


sensorium:
Hi Thomas,

it sounds like a well-reasoned change with lots of advantages.  If it increases the processing efficiency, great.  

Your contributions open the possibilities for a wider community to work with the software on newer platforms.  I initially envisaged Mozzi as a community project and licensed it as gnu v3.0.  However, due to a lack of community contribution and an episode when a company sold Mozzi software without acknowledging where it came from, I changed to a CC-share alike-noncommercial license.  This was so I could negotiate with people who planned to make money from Mozzi, while I put in a lot of solo unpaid work.

Do you have any thoughts about licensing and attribution?  I am open to any ideas about it.  I would like to encourage more community contribution, but would also like to protect my own work from ill-spirited commercial exploitation.  I usually suggest to people who make contact about using Mozzi in commercial products that the first 100 units sold are royalty-free, to help small scale artisan production, then for each unit after that perhaps 1% of the sale price would be paid as a royalty - to me, so far, as the main author of Mozzi.  To date, only one person has responded and the royalties have totalled less than an afternoon's pay over the last 5 years.

Well, no clear message there from me, but on the technical side, your collaboration is welcome.  Let me know what you think about the licensing issues.

Regards,
Tim

tfry-git:
> it sounds like a well-reasoned change with lots of advantages. If it increases the processing efficiency, great.

Glad you like it. I'm going to turn this into a "real" patch, then, shortly.

On the matter of attribution: If I get a one-time mention in the announcement of the next release, or something like that, my desire for fame is going to be absolutely satisfied.

Regarding licensing: I usually license my projects under GPL 2/3. Of course my contributions to existing project such as yours are available under that project's license. I hereby let it be known, that Mr Sensorium may re-license my past and future (until further notice) contributions to Mozzi in any way he sees fit, under the sole condition that whatever version of Mozzi they were _first_ included in will remain available under whichever distribution licence(s) said version of Mozzi was available at the time of the contribution.

Regarding financial compensation: That's roughly my experience for my "big" project, as well (http://rkward.kde.org). I suppose you'll have to live on positive user feedback... I do not raise any claim to any share of any financial compensation you may receive in the future. Although, in case the project does turn you into a tech billionaire, after all, I wouldn't mind if you pass on a few breadcrumbs ;-).

Regarding community contribution: Yup, that is a difficult beast. One thing that appears to be lacking is a "common" platform for discussing, sharing, and managing contributions to Arduino-projects, in general. Github is kind of nice, technically, but ... just so fragmented. It doesn't seem good at creating a community. 

One thing that _maybe_ the Mozzi project could do is to provide a vendor-neutral forum or mailing list. At least for me, having your discussion in a google group, simple meant an entry barrier. Usually I would have taken my ideas, there, first, but didn't want to sign up. Well, I already had a github-account, so I used that, instead, but usually I am a bit reluctant about signing up to yet another internet service, and I suppose _some_ others will feel the same, so there's a danger of turning people away before they have actually decide to step in. Personally, I love mailing lists (can you tell I'm old?). Perhaps take a look a FreeLists. All that said, however, _if_ that helps, it won't help _much_, so may not be worth the trouble.

tfry-git:
Ok, so here's the "real" version of the patch. I've split it in two logical parts, in case the second part should need reverting (I don't see why it should, but just in case).

Part 1: Removes timer 0 code from MozziGuts.cpp, calling updateControl() from audioHook(), instead. Updates API documentation. Does not remove utility/TimerZero.*, in case users have already started using that (or will start using it now that Timer 0 is "free").

Part 2: Removes ATOMIC_BLOCK() calls everywhere, as there is no longer any danger of updateControl() stepping in, while updateAudio() is in the middle of some function assessing the same variables. I have done only very basic timing on this part, and only on STM32, but in my "real world" sketch, it reduced the time needed for 32768 iterations of updateAudio() from 375ms to 370ms. Not a huge deal, but it seems reasonable to reap this additional benefit, if it does not turn out to cause any trouble.

What this PR does _not_ address is updates to non-API documentation. "Under the hood" is definitely a bit off, now, and there are various mentions of (lack of) Timer 0 all over the place. I had intended to take care of these as well, but - as far as I understand - I cannot access these in source form, and thus cannot provide a patch. Let me know, if there is a way for me to help with that.

sensorium:
Thanks for this Thomas - I'm running behind with things at the moment but I look forward to merging when I have time to look at it properly.
Actually, I'll send you an invitation to collaborate so you can merge your own changes when you're happy with them.
Also, thanks for your offer to update the non-API documentation.  I'm in the process of migrating the website framework from octopress 2 to 3, and I'll see if I can work out how to get it all editable in git.
Apologies for my delays, I appreciate the time you're putting in.

tfry-git:
Many thanks for the push access! I'll try to use this responsibly. I think I'll probably merge this PR, shortly (after taking another good look, myself), as it will help me with the ESP8266 port. But in general, I'll continue to ask you for feedback on any changes. And no worries, I'm not in a hurry. I always have enough side-projects running to keep me busy... Keep me posted on what I can do WRT documentation updates.

PR #39, opened 2018-01-18
Fixed: ADSR level was not re-set to 0 when noteOn() is called, while …

tfry-git:
…the envelope is still "playing".

---

Sorry to be flooding you, but here's another pull request. This one should be easy, though. Problem is that the ADSR level is not re-set to 0, when calling noteOn() while the env is still in "playing" state, e.g. when re-cycling an effect envelope.

### DIFF
--- a/MozziGuts.h
+++ b/MozziGuts.h
@@ -24,6 +24,7 @@
 #include "hardware_defines.h"
 #include "mozzi_analog.h"
 
+#if not defined (CONTROL_RATE)
 /** @ingroup core
 Control rate setting.
 Mozzi's CONTROL_RATE sets how many times per second updateControl() is called.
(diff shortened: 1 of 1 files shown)


sensorium:
Hi Thomas,

I suppose I did it that way as a "feature" rather than a bug, to avoid sudden jumps in envelope level, so the sound level or parameter continues from its current position when re-triggered... would have to listen to some tests to see what's preferable, and perhaps add a flag or function for the desired behaviour...

Tim

tfry-git:
Sure, your call. I took "THis will restart the ADSR no matter what phase it is up to" to mean the current behavior is not intended. But I'll also be happy with a separate "hard reset" function/parameter.

PR #40, opened 2018-01-19
New attempt: Allow to force an ADSR envelope back to zero on noteOn().

tfry-git:
On further thought, you are absolutely right: Going back to zero level on noteOn(), _unconditionally_, is not desirable. For "normal" note amplitude envelopes, it will be best to continue from the current level, if a note is re-triggered while still playing (esp. during release). However, for other use cases of envelopes (effect envelopes), going back to zero may be much more desirable.

So the new attempt leaves the choice to the user, with explicit documentation.

### DIFF
--- a/ADSR.h
+++ b/ADSR.h
@@ -170,10 +170,14 @@ public:
 
 
 
-	/** Start the attack phase of the ADSR.  THis will restart the ADSR no matter what phase it is up to.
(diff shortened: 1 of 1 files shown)


felippeduran:
I've tried this fix and it works like a charm. I've been stuck for a few days looking for answers in the forum regarding this behaviour until I ended up here and it is indeed a bit misleading.

It seems that this pull request also fixes issue #39.

PR #42, opened 2018-02-24
WORK IN PROGRESS: Port to ESP8266

tfry-git:
This one is _not_ ready to be merged, but to let you know of the work being under way (**and to invite potential contributors to the port**).

Summary: Adds support for the ESP8266.

Status: Compiles, but as I found out, too late, it **cannot** be made to work on any of the ESP8266 boards for hardware reasons. Waiting for another ESP8266 board to arrive in the mail, before I can proceed. As such, the port works **in theory**, but that theory is entirely untested.

Issues:
- The ESP8266 only has one ADC pin, and nobody seems to have figured out, how to read it asynchronously. Analog read functions are unimplemented, but mozziAnalogRead() forwards to analogRead().
- Timers usable for updateControl() only have millisecond resolution. Not ideal. ESP8266 would profit from https://github.com/sensorium/Mozzi/pull/38 .
- atomic.h is not available, either. Crudely disabled for now. This will either need some more effort, or, again, https://github.com/sensorium/Mozzi/pull/38
- For the ESP platform, it will probably be **much** more performant, if wavetables / samples / oscillators could be stored in RAM, as flash is connected externally, while RAM is - comparatively - abundant.


tfry-git:
A small update and a question:

The update: I thought of another way to push out PDM coded audio, and now I can use Mozzi on my ESP8266. The examples I tried so far are finally working, but it did run into a bunch of subtleties along the way, that did not cause noticeable problems on the other architectures, but cause crashes on the ESP8266 (division by zero is one example).

The question: As it turns out keeping even small samples / wavetables in flash is a complete no-go for the ESP8266. As written before, it makes up for this deficiency in RAM, and so that's a reasonable alternative. __But__ this means touching a whole lot of code. The one half of it is the wavetable definitions. For those I created a new #define "CONSTTABLE_STORAGE" that will put these into RAM on the ESP8266, while keeping them in flash, elsewhere. Somewhat ugly, but unavoidable.

Now, the second half of this is the read accesses in Oscil.h, Sample.h, etc. So far these use pgm_read_byte_near(). However, according to my reading, that's mostly a historical cludge, while nowadays, you'd put data in flash using the "__flash" qualifier, and access them like a regular array.

Thus: While working on this, should I switch to "__flash" and array access, or should I keep things as they are (with indirection over a preprocessor define)?

sensorium:
Hi Thomas,
more great work!

>pgm_read_byte_near() .... mostly a historical cludge, while nowadays, you'd put data in flash using the "__flash" qualifier, and access them like a regular array.

Can you give a link about this?  For education... thanks.

>should I switch to "__flash" and array access

Sounds like a sensible update to me.

tfry-git:
> Can you give a link about this? For education... thanks.

Ha! I wish I could seem to find a decent write-up. Anyway, from what I gather from various forum posts, here and there, the __flash keyword is not part of the C++ standard. It has long been supported by some compilers, but not GCC. It was finally [added in GCC 4.7.0,](https://gcc.gnu.org/gcc-4.7/changes.html) however, i.e. mid 2014, and this addition is what has changed.

Documentation on the usage of the keyword is at https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/Named-Address-Spaces.html . However, the most important bit is easy to miss: This is formally a qualifier, i.e. part of the data type. Thus, while "const __flash int8_t*" it still a 16-bit pointer, as before, the compiler now knows that it is a pointer to something inside the flash region. Therefore, it can resolve it using regular addressing, without the need for special access functions.

In the context of Mozzi, the best thing about it is that it could make it trivially(?) easy to allow higher bit-resolutions in Oscil, Sample, and WaveShaper, without adding much complexity. (By re-basing these on a simple template class, each). While this is not something I'm presently missing, but I could imagine this to become of interest for some users on the more resourceful CPU architectures.

tfry-git:
> the __flash keyword [...] was finally added in GCC 4.7.0, however, i.e. mid 2014, and this addition is what has changed.

Oops, forget about it. The __flash keyword is supported in the C compiler, but _not_ in the C++ compiler, for whatever reason. So it's off-limits, sorry about the red herring.

I suppose, instead, ``pgm_read_byte/word/dword_near()`` could be wrapped inside an (inline!) template function that would allow to write ``CONSTTABLE_READ<int8_t>(table, position)``, instead of ``(int8_t) pgm_read_byte_near(table + position)``. But I guess that is something to worry about, if and when actually implementing support for different data types in Oscil and friends.

tfry-git:
Ok, I now consider this port "ready". I would appreciate if you could take at least a cursory look for any obvious omissions or unjustifiable ugliness.

One of the things is that I now changed `const int8_t CONSTTABLE_STORAGE bla` to `CONSTTABLE_STORAGE(int8_t)`. This would allow to switch to the __flash keyword more easily, if it ever becomes available. Or it could be used for fancy tricks such as runtime modifiable tables, now that they are already kept in RAM on the ESP8266. However, that's all totally optional, and I can revert it to the previous state, if you prefer.

tfry-git:
Ok, thanks for the thumbs up. I'll merge this, then.

For what it's worth, I did some cursory testing on 328P, but I certainly did not cover all areas, in particular no HIFI, and stereo at all. I did note that at least on Detuned_Beats, there is some clicking due to buffer underruns, unless removing one of the oscillators from updateAudio(). That does not seem all new, though, it also happens at 0a984fff30c3073e5d395f1517bd4bb3fc930aec , so I hope it's not my fault... (and I did not heed the advice to switch to -O2, either).

Issue #43, opened 2018-02-27
RFC: Scaling output samples when switching audio configuration or CPU


tfry-git:
I don't know how many people are playing / will be playing with Mozzi on different CPUs, but one of the annoying aspects while cross-testing is that different ports expect a different audio resolution to be returned from updateAudio(). The same problem applies when switching between available audio configurations on the same CPU.

An example written for StandardPlus (8.5bits or so) will be somewhat quiet on the STM32 (10 bits), very quiet on the Teensy (12 bits), definitely too quiet in HIFI mode (14 bits), and barely audible on the upcoming ESP8266 port (16 bits). That's rather sad, as everything else in a typical Mozzi sketch _can_ be expected to work, unchanged. You would not believe how many times I went hunting for "what did I break now" while cross-testing, when the whole problem was that I forgot to adjust updateAudio() to my current configuration. I wonder how many users will experience similar woes when trying something new. Another symptom of this problem is "duplicate" examples for standard and HIFI modes.

So: Wouldn't it be cool, if Mozzi could do the necessary scaling all automatically? (Yeah!) But how?

1. A first, naive approach would be to make all ports assume the value returned by updateAudio() are 8 (or 9) bit resolution, these could then be scaled up to the output bits inside audioHook(), for those ports/configs that support more bits. But this approach has severe problems:

- Breaks existing sketches using higher resolution in updateAudio()
- Adjusts the volume, but limits the effective audio resolution to 8 bits, where some ports could offer much better quality.

2. A second approach would be to make all ports assume samples scaled to - say - 16 bits returned by updateAudio(), and scale down from this, automatically, as appropriate. As the intermediate resolution is high, each port can support maximum quality. However, this time, the problems are:

- Breaks the vast majority of existing sketches returning 8-9 bits in updateAudio()
- Places an additional (small) processing requirement on the lower end CPUs, where processing power is already most scarce.

3. A third approach will be to use an opt-in "best practice" solution to automatic scaling. This would specify the number of audio bits generated, and auto-magically scale those to the appropriate output resolution. To elaborate, this would rest on macros like this (where AUDIO_BITS is the desired output resolution):

```
// Assume that "value" is scaled to "bits" resolution, and scale it to the output resolution appropriate for the current configuration
// As long as bits is a compile time constant, the compiler should be smart enough to optimize this to a static shift operation (or nothing)
[...]

PR #44, opened 2018-03-18
Support for Teensy LC, 3.5, 3.6 and AVR 8Mhz

adrianfreed:
Added support for Teensy 3 LC, 3.5, 3.6.
Corrected bad #ifdef added with STM32 support that broke Teensy support. Includes new support for 8MHz AVR tested on Lilypad USB.


README.md
```diff
@@ -125,6 +126,7 @@ If you still need more speed, Arduino 1.0.5 produces slightly faster code.
 
 ## Caveats and Workarounds
 
+* AVR
```
tfry-git: In the current version of the doc, this section is no longer AVR specific. (Yes, millis() and delay() are available on AVR, too, today, but it is not smart to use them - delay in particular - on any architecture)

README.md
```diff
@@ -192,21 +194,21 @@ real DAC. Should probably run on any other board supported by [STM32duino](https
 - Note that AUDIO_INPUT and mozziAnalogRead() return values in the STM32's full ADC resolution of 0-4095 rather than AVR's 0-1023.
 - twi_nonblock is not ported
 
-### Teensy 3.0/3.1
+### Teensy 3.1/3.2/3.4/3.5/LC
```
tfry-git: What about Teensy 3.0? No longer supported?

MozziGuts.cpp
```diff
@@ -244,6 +245,76 @@ ISR(ADC_vect, ISR_BLOCK)
 	adc_count++;
 }
 #endif // end main audio input section
+#endif
+
+
+#if IS_SAMD21()
+#define AUDIO_BIAS ((uint16_t) 2048)
+
+static bool tcIsSyncing()
```
tfry-git: Can you explain a bit, what these tcXXX() functions are doing (add comments and / or use a more explanatory naming). I have half a mind to go try to clean up all the hardware-specific \#ifs one of those days, but for that purpose it would be really important to still be able to understand what's going on.

MozziGuts.cpp
```diff
@@ -346,20 +417,53 @@ void audioHook() // 2us excluding updateAudio()
...
+        
+     void samd21AudioOutput()
+    {
+        
+#if (USE_AUDIO_INPUT==true)
+        adc_count = 0;
+        startSecondAudioADC();
+#endif
+        
+        analogWrite(DAC0, (int)output_buffer.read());
```
tfry-git: I suggest using a define in place of DAC0 (even if it may not be user configurable). That will be more in line with the other architectures. Also, this function looks pretty similar to the Teensy3 one. Could they merged?

Oscil12.h
```diff
@@ -0,0 +1,325 @@
+/*
```
tfry-git: Higher bit oscillators make a lot of sense on the 32 bitters, but please split them out of _this_ PR, as they are only partially related.

(And I think we'll probably want a more generic - templated based - solution, then.)

README.md
```diff
@@ -192,21 +194,21 @@ real DAC. Should probably run on any other board supported by [STM32duino](https
...
 
 Extra libraries required for use withTeensy 3.0/3.1:
-
+These are included in the standard Teensyduino install
```
tfry-git: Does that mean Teensy users will not have to worry about installing these, today? Then I'd suggest removing the notice. Or is that only for a subset of Teensy's, or necessary in some circumstances?

README.md
```diff
@@ -192,21 +194,21 @@ real DAC. Should probably run on any other board supported by [STM32duino](https
...
 
-- Audio output is on pin A14/DAC, in STANDARD or STANDARD_PLUS audio modes, These modes are identical on Teensy 3.0/3.1, as the output is via DAC rather than PWM.
+- On Teeensy 3.1/3.2/Audio output is on pin A14/DAC, in STANDARD or STANDARD_PLUS audio modes. These modes are identical on Teensy 3.0/3.1, as the output is via DAC rather than PWM.
 - Output is 12 bits in STANDARD and STANDARD_PLUS modes, up from nearly 9 bits for Atmel based boards. HIFI audio, which works by summing two output pins, is not available on Teensy 3.0/3.1.
-- #include <ADC.h> is required at the top of every Teensy 3.0/3.1 sketch.
+- #include <ADC.h> is required at the top of every Teensy 3 sketch.
 - The examples come with this commented out, for Arduino compatibility.
-- Serial baud rate for monitoring in the IDE needs to be set to 9600 to work with Teensy 3.0/3.1. This slow rate can cause audio glitches.
-- twi_nonblock code by Marije Baalman for non-blocking I2C is not compatible with Teensy 3.0/3.1.
+- Serial baud rate for monitoring in the IDE needs to be set to 9600 to work with Teensy 3.1/3.2. This slow rate can cause audio glitches.
```
tfry-git: Again, what happened to Teensy 3.0?

MozziGuts.cpp
adrianfreed: They are the timer functions analogous to the Teensy ones which were built into a nice library. I haven't been able to navigate the SAMD21 universe well enough to see if there is a good library. The problem with the code I added is that I picked a specific timer which may conflict. I will ask Dave Mellis who is managing the conventional use of these across the libraries.

MozziGuts.cpp
adrianfreed: They could be merged but this way there is room for I2S support which is likely different between SAMD21 and Teensy.

README.md
adrianfreed: The teensyduino installer lets you choose which libraries to install, so conceivably somebody might uncheck the box for these libraries and not know how to proceed. This library dependency issue is not handled well by Arduino at this point...

Oscil12.h
adrianfreed: Well they make sense on AVR with Timer4 (Leonardo, Lilypad USB) with 10-bit (and maybe 11) PWM. Because they are in a separate header file there is no cost to include them.  I am not sure how to templatize this well as you have to change the PROGRAM call according to byte or word lookup.

README.md
adrianfreed: Yes, Documentation corrected now

MozziGuts.cpp
adrianfreed: It may be that DAC0 is now the "Official" Arduino way of naming the DAC pin. I am loath to create yet another name. None of these names work very well anyway because it is called A0/D0/~ for example on Arduino Gemma. The ~ is the most helpful actually.

Oscil12.h
adrianfreed: I just confirmed that the FM example runs fine with a 16-bit carrier table on Lilypad USB (8MHz) and Teensy 2.0

README.md
adrianfreed: Fixed

MozziGuts.cpp
tfry-git: Just add a comment along the lines of "these functions (tcXXX()) are used to run a timer at AUDIO_RATE, to copy samples for the output buffer to the DAC on SAMD21". (If I have understood their purpose, correctly). I.e. document the _why_ of these functions, somehow.

MozziGuts.cpp
tfry-git: What I means is

```
#define AUDIO_CHANNEL_1_PIN DAC0
```

somewhere. That is used consistently _inside Mozzi_.

Oscil16.h
```diff
@@ -58,7 +58,7 @@ char2mozzi.py infilename outfilename tablename samplerate
 */
 //template <unsigned int NUM_TABLE_CELLS, unsigned int UPDATE_RATE, bool DITHER_PHASE=false>
 template <uint16_t NUM_TABLE_CELLS, uint16_t UPDATE_RATE>
-class Oscil12
+class Oscil16
```
tfry-git: Yes, going to 16 bits makes more sense than 12 bits, _but_, again, please just split it out into a separate pull request. I think we can come up with a yet more generic solution, but that's simply a separate discussion, and should not hold up the extended Teensy 3.x and SAMD21 support.

Oscil12.h
tfry-git: To clarify, I am absolutely _not_ opposed to a change along those lines, just please make it a separate pull request.

As for the templating, I think the solution will be to templatize CONSTABLE_READ(), too. But again, separate PR, please.

(And while doing this, it _may_ - or may not - make sense to also take a look whether Sample and Oscil can be made to share a common template.)

MozziGuts.cpp
adrianfreed: OK

MozziGuts.cpp
adrianfreed: ok

adrianfreed: Since you have a platform specific place for such defines, why not put the DAC range in there too? e.g. DACRANGE 4096, 2048, 256. Then the output calls can have something like * DACRANGE/256  (or DACRANGE/ 65536 with a 16-bit table). These would be optimized into shifts by most compilers...

tfry-git: Yes, you are right, something along those lines will be needed, consistently for all platforms. You may want to comment on my proposal at https://github.com/sensorium/Mozzi/issues/43 . Not sure this is the best way to do it, hence the ticket for discussing it, first.

tfry-git:
Looks good to me, now, but I'll try to do some testing in the next few days (on STM32, ESP8266, and 328P). Which platforms did you test, yourself?

@sensorium As this adds support for 8MHz AVRs (seemingly trivial), I'm curious: Did you simply never get around to do this, or was this tried and discarded, for some reason?

MozziGuts.h
```diff
@@ -76,18 +76,18 @@ x...11.......Arduino Mega  *broken since Jan 2015 \n
...
+On Teensy 3.* STANDARD and STANDARD_PLUS are the same, providing 16384Hz sample rate and 12 bit resolution on pin A14/ADC.
+The Teensy 3.* DAC output does not rely on PWM.
 */
 
 
 /** @ingroup core
 Used to set AUDIO_MODE to HIFI.
 
-HIFI (not for Teensy 3.1)
+HIFI for AVR (not for Teensy 3.* and ARM)
```
tfry-git: Ha! Just found another small reason to nag ;-)

STM32 actually has support for HIFI mode (and is an ARM). ESP8266 does not, and is not an ARM. Same a few lines below.

extras/python/sin1024_int8.py
```diff
@@ -1,40 +0,0 @@
-## generates a sin-shaped table with values -128 to 127
```
tfry-git: Accidentally removed while splitting out the Oscil16-changes?

MozziGuts.h
adrianfreed: oh whoops. This is gonna get complicated. You are doing PWM and HIFI on STM32 but most of the STM32 boards I have seen have a pair of 12-bit DAC's. It's gonna be a conditional compilation zoo in the near future.

extras/python/sin1024_int8.py
adrianfreed: whoops

MozziGuts.h
adrianfreed: Incidentally, most of the libraries are getting this wrong because Arduino only just sorted out some conventions about board #defines (and Teensy is not "on board" yet). Teensy code conditions on the particular chip for each Teensy board rather than a board description. So far this works because Paul is the only one employing those chips. Using an architecture description like ARM or STM32 or SAMD may be too broad. My favorite way you have addressed this so far is your list of tested boards... I wish I had time to test all the Mozzi examples on all the different boards I have (30+).

adrianfreed:
I am testing on Teenys 2, Teensy 3.2, LC and Gemma M0 and Lilypad USB and occasionally on Adafruit playground express and Gemma. I teach a class on this in two weeks, so students will presumably stress test some more.

MozziGuts.h
tfry-git: > oh whoops. This is gonna get complicated. You are doing PWM and HIFI on STM32 but most of 
> the STM32 boards I have seen have a pair of 12-bit DAC's. It's gonna be a conditional compilation 
> zoo in the near future.

Yes, all but the cheap STM32's have built-in DAC, and one day we'll want to support those, natively. Add to that I plan on adding PDM output (more CPU intensive, but potentially much higher quality) for STM32, too, and all those \#ifdefs are pretty messy, already. That why I insisted on you adding those comments! I really hope we can think up a sane way of organizing this zoo, one day...

sensorium:
HI Thomas and Adrian,

it's great to see your contributions making Mozzi more useful and robust.

A few days late, but in response to:
>support for 8MHz AVRs
>Did you simply never get around to do this, or was this tried and discarded, for some reason?

It's a long time ago and I think I just didn't get around to it.  I vaguely recall early testing, when Oscil frequencies were incorrectly calculated on an 8MHz AVR and I was only focussed on getting it to go, rather than porting.

btw, apologies for my apparent lack of attention here at the moment - My local development environments have collapsed since migrating to a new computer.  I'm struggling with updating the Mozzi website framework, including making the documentation editable on Github.  And also struggling with how my local Mozzi version relates to the master branch on Github - there are half-made changes at home and on Github which need cleaning up.  I'm somewhat frozen on both fronts so thanks for your patience as I try to get my act together...  (I don't even know the proper place to put this).

sensorium:
> I seem to recall having a problem getting the correct pwm carrier frequency and counter range with 8MHz, but I probably have missed or forgotten something...

PR #45, opened 2018-03-19
Allow to skip ADC setup.

tfry-git:
E.g. for compatibility with certain libraries, or to enable techniques such as
analog port multiplexing.

Note that this is a run time solution. On the positive side, users can select to
skip ADC setup per sketch, without making any changes in the lib. On the negative side
(compared to a compile-time switch), this solution does not save any allocated RAM, and
the ISR routine is still defined on platforms that define it.

---- For some more background:
I had stumbled across this before. My present issue was that I tried to get a resistive touchscreen to work on an STM32F103C8T6, but this turned out to be impossible in conjunction with Mozzi. Use-cases like this are going to become more widespread on the more powerful platforms. But even on 8bit AVRs, there are cases, where users will need custom analogRead()s - e.g. from the PotMUX library.

### DIFF
--- a/MozziGuts.cpp
+++ b/MozziGuts.cpp
@@ -680,9 +680,9 @@ static void startControl(unsigned int control_rate_hz)
         update_control_timeout = AUDIO_RATE / control_rate_hz;
 }
 
-void startMozzi(int control_rate_hz)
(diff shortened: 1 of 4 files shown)


tfry-git:
Closing in favor of #211 , which adds a compile time option to this effect.

Issue #47, opened 2018-03-21
reading some wavetables from SD card


PantalaLabs:
Hello guys

I'm developing an oscillator that have many wavetables that dont fit into board memory.
I cna load then from SD card because this operation is made in "stand by" mode , so I can stop all oscillators and change my wavetables .
But , I cant figure out how to read a new wavetable that wasnt loaded at compile time.
Any clue ?

Thank you!!

tfry-git:
This is not currently possible with Oscil (at least not on AVR), because that assumes that the wavetable is stored in flash memory. Instead, when loading from SD card, you'll have it in RAM. If you feel adventurous, you could try modifying the CONSTTABLE_READ macro (globally, or - safer - only in Oscil.h) to simply read a value from a plain RAM address (as done for ESP8266, in hardware_defines.h).

There are currently some ideas for extensions to Oscil, and possibly an option to this effect could be implemented as well (but I'm not yet sure, whether / how that can be done).

PantalaLabs:
Thank you for the explanation.
I'll study the macro and if I found a solution , I write a topic here .

adrianfreed:
You might be able to add other methods that handle a non-const table pointer type?
Then you wouldn’t need a macro, probably just an array index.

tomcombriat:
Hi, sorry for the delayed response.

Have you tried? I had the feeling #48 would allow that but I have to say that I did not try or looked closely into it nor was implicated in it, but as it can read from RAM now…

PR #48, opened 2018-03-22
Replace CONSTTABLE_READ with a template-based (inline) function that …

tfry-git:
…will

- work with addresses in both RAM and flash
- automatically pick the correct pgm_read_XXX_near variant() based on data type

The main point is to allow implementing Oscillator and Sample template classes that
can work on custom data types (int16_t, float, etc.), and also allow to work with
samples stored in RAM.

See also https://github.com/sensorium/Mozzi/issues/47
and the related discussion in https://github.com/sensorium/Mozzi/pull/44

@adrianfreed So this is what I think should enable you to implement higher resolution oscillators as a template class. What do you think?


adrianfreed:
I like this a lot. I would suggest adding the case of 8byte sizeof for double. It took a while but in a few years we will have a lot of native float and double CPUs running Arduino code. What I like about your template is that is generalizes Oscil to the point it can be extended to do granular synthesis for example. I think if I can templatize the return type of the index functions and the private table pointer we will have a powerful "separation of concerns" between the storage and the indexing and phase computations.

adrianfreed:
My fork has your template and now the higher resolution template. I haven't tested it thoroughly yet. . I am not happy with the design. I default to const int8_t for backwards compatibility. It seems to me that the table size can be inferred at compile time. For that matter can't we also infer the base element type from the  argument pointer to the table? If this is true it would simplify the API (and break the current one).

tfry-git:
I did not add sizeof(T)==8, so far, because there is no pgm_read-variant for that. But it should be possible to do using two reads. However, I also just realized that my macro will do a wrong cast when reading floats. I'll have to look into the correct magic for that...

About automatic type inference, I usually shy away from that if I can, but that's probably, because I don't understand it well enough. I don't think table size can be inferred in all cases. Consider the guy talking about reading wavetables from SD card. Suppose he wants to use a dynamically allocated buffer for that. That would leave only a naked pointer to the compiler. Not sure, why it would actually matter, though? Can you post a minimal pseudo-example?

So far I had assumed that you'd be creating an all separate template, and the current Oscil would be re-defined a partial specialization of that. Not sure, whether that is the only or best way to go, but I would argue that keeping the current API intact is a hard requirement.

adrianfreed:
I have no problem with keeping the current API intact. The kind of generalization I am talking about (with granular synthesis) expands the scope to the point that even the name Oscil is too narrow. It is  really about efficient tabulated function evaluation with high resolution phasors. I will reflect on this and experiment for a while before producing a pull request. It's an interesting return for me to questions of expressivity and simplicity for these signal processing and synthesis functions.

tfry-git:
Ok, understood. Just in case, I'll also point you to the Sample-class. Not sure, this can reasonably be merged into the same code-base, but from a cursory look, probably it can. Also the optional interpolation will make a lot of sense for Oscillators, too, when value resolution is (much) higher than temporary resolution.

tfry-git:
@sensorium Ok, this one has become more complex than I would have hoped. I _think_ I got it now, and testing is successful. However, I would not be all that surprised, if this needs tweaking for some architecture in the future. To support this, I'd like to add a small test-case somewhere (which I have used for testing locally: It simply initializes various data types in flash and RAM, then prints them to Serial to make sure they are retrieved, correctly). But - where to put it? This is not really suitable as an "example", as it does not do anything a user would have to worry about.

(For the time being, here it is:)
```
#include <MozziGuts.h>

CONSTTABLE_STORAGE(byte) a[] = { 1 };
CONSTTABLE_STORAGE(int8_t) b[] = { 2 };
CONSTTABLE_STORAGE(uint8_t) c[] = { 3 };
CONSTTABLE_STORAGE(int16_t) d[] = { 4 };
CONSTTABLE_STORAGE(uint16_t) e[] = { 5 };
CONSTTABLE_STORAGE(int32_t) f[] = { 6 };
CONSTTABLE_STORAGE(uint32_t) g[] = { 7 };
CONSTTABLE_STORAGE(float) h[] = { 8.8 };
CONSTTABLE_STORAGE(double) i[] = { 9.9 };

byte ram_a[] = { 1 };
int8_t ram_b[] = { 2 };
uint8_t ram_c[] = { 3 };
int16_t ram_d[] = { 4 };
uint16_t ram_e[] = { 5 };
int32_t ram_f[] = { 6 };
uint32_t ram_g[] = { 7 };
float ram_h[] = { 8.8 };
double ram_i[] = { 9.9 };

void setup() {
  Serial.begin(9600);
}

void loop() {
   Serial.println("Flash reads");
   Serial.println(FLASH_OR_RAM_READ(a));
   Serial.println(FLASH_OR_RAM_READ(b));
   Serial.println(FLASH_OR_RAM_READ(c));
   Serial.println(FLASH_OR_RAM_READ(d));
   Serial.println(FLASH_OR_RAM_READ(e));
   Serial.println(FLASH_OR_RAM_READ(f));
   Serial.println(FLASH_OR_RAM_READ(g));
   Serial.println(FLASH_OR_RAM_READ(h));
   Serial.println(FLASH_OR_RAM_READ(i));
   Serial.println("RAM reads");
   Serial.println(FLASH_OR_RAM_READ(ram_a));
   Serial.println(FLASH_OR_RAM_READ(ram_b));
   Serial.println(FLASH_OR_RAM_READ(ram_c));
   Serial.println(FLASH_OR_RAM_READ(ram_d));
   Serial.println(FLASH_OR_RAM_READ(ram_e));
   Serial.println(FLASH_OR_RAM_READ(ram_f));
   Serial.println(FLASH_OR_RAM_READ(ram_g));
   Serial.println(FLASH_OR_RAM_READ(ram_h));
   Serial.println(FLASH_OR_RAM_READ(ram_i));
   delay (2000);
}
```

sensorium:
How about a new folder extras/tests?

sensorium:
Just some slight anxiety as the code develops ahead of my ability/time to follow it at present - are the changes going to make it more difficult for learners to get a handle on how Oscil (or Sample, if they're combined) works?  Will there be default template parameters that can be ignored by beginners?  
Part of the reason for Sample being seperate (for example) is that it's easier conceptually than having an all-purpose oscillator with lots of buried functions, requiring in-depth knowledge.  Learners already complain that Mozzi has a "strange syntax", since templates are unfamiliar to many...

tfry-git:
Understood. Well, this PR in itself should not change anything. Regular users will never get in touch with this. (And indeed it looks rather terrible, but the whole point is to be able to re-write pgm_read_XYZ_near() with a more templating friendly FLASH_OR_RAM_READ<const int8_t>() (or whatever other type).

But of course, yes this PR is here to enable a more flexible implementation of Oscil and/or Sample. Regarding this, let's see what @adrianfreed can come up with. I think, quite naturally, this will introduce new choices, importantly between staying a 8bit integers or going upwards, and between staying in flash, or going to RAM (which would enable playing samples from SD, for instance).

But it should also be no problem to keep Oscil and Sample fully intact, only I suppose they might internally become an "alias template" i.e. essentially a typedef for a special flavor of the new more flexible template. Similarly, I suppose it should be no problem to provide aliases such as Oscil16 or RAMOscil16 for some of the more obvious alternative specializations users might want to try.

adrianfreed:
The currently envisaged changes don't make the API any harder to understand.  const/program memory is a complexity bump in the road for people to cover eventually so abstracting away the details in a portable way only helps the beginner keep their eyes on the prize. I will teach shortly with things pretty much as they are as I don't have a good sense yet of what difficulties people meet
when the encounter the code. My sense is that C++ isms are not easy to grasp from examples and that Arduino libraries which are becoming increasingly sophisticated in their use of C++ are quite a hurdle for beginners.  The Mozzi library is interesting to me precisely because it appears to be possible to grow it to take advantage of faster processors without losing all the efficiency programming for the 8-bit machines.  This is why I tweaked for the 8MHz in the low end. Many of the wearable boards I teach with use 32u4s at the lower speed for LiPo voltage levels. I know it's possible to combine sampling and oscillation because Max/MSP, for example, already does this (in an interpreted environment). They use a storage/access separation of concerns. Their storage abstraction is called a "buffer."   I have no idea if doing this would result in something harder or easier to understand for beginners. The diversity of the Arduino community is making it hard to know what a beginner might know any more. They are all climbing the learning curve so fast these days.

adrianfreed:
Incidentally, I am a big fan of templates for efficiency programming and their use is long overdue  in embedded systems programming. I jumped on them in the early days and wrote a paper with a colleague on  how they can also improve the legibility of sound synthesis code (https://pdfs.semanticscholar.org/ddcb/697c482d8cf329963e7d1d72bf31a2ea5413.pdf). It may be interesting to think of Mozzi as leading a positive move towards them rather than resorting to an apologetic stance.

tfry-git:
@sensorium You voiced scepticism. Is it ok for me to merge this one?

sensorium:
Yep, merge when you're ready.

PR #49, opened 2018-05-09
Adding possibility to use MCP4922 with Mozzi

tomcombriat:
This enables possibility to use MCP4922 SPI DAC with Mozzi. If wanted, this can be chosen in mozzi_config.h. This will print a warning message at compilation. This allows a true DAC output, even with slow running devices, hence a lesser need for post filtering.

A port can easily be made for other MCP DAC, but I do not have the hardware to test them.

SPI and MCP49XX libraries has to be installed to use this feature.


tfry-git:
Hi!

Thanks a lot for your patch. Supporting external DACs is certainly an interesting extension.

Unfortunately, your patch includes a huge amount of whitespace changes. I assume these happened in your editor, automatically. But the result is that it is near impossible to see what your patch is actually changing. Please try to fix this, somehow, possibly by using a different editor, and / or configuring your editor to use tabs instead of spaces for indentation (the indentation style throughout the Mozzi library). Please look at the resulting diff!

I suggest to back out the change to Oscil.h. I don't think it does what you mean it to do. Higher resolution Oscillators are something we are actively thinking about (esp. @adrianfreed), but we really need a solid solution, here, which should be discussed, separately.

I do understand you have only a certain type of DAC to test with, but I suggest to keep the macro- and file-naming generic. Others are likely to expand on your code, eventually, and then it seems rather likely that all external DACs (or at least those connected via SPI) will need very similar code.

tomcombriat:
Hi Thomas,

Thanks for taking time to give suggestions!

I did a few changes in order to take into account your advices:
  - I backed up Oscil.h to original: I was not intending anything but this was a unwanted modifications that remained from previous tests
  - I changed the name of the patch in order to make it more generic for all SPI DAC from the MCP49XX series. It is now easy to change the type of DAC used in AudioConfigMCP49XX.h
  - I cleared the unwanted whitespaces (sorry for that!).

If you have any further suggestions please tell me.

AudioConfigMCP49XX.h
```diff
@@ -0,0 +1,17 @@
+#ifndef AUDIOCONFIGMCP49XX_H
```
tfry-git: I suggest naming this file AudioConfigExternalDAC, so that it will be easier to add further DACs in the future (without creating a separate file for each).

MozziGuts.cpp
```diff
@@ -35,6 +35,9 @@
 #include <uart.h>
 #include <Ticker.h>
 #endif
+#ifdef __DAC_MCP__
```
tfry-git: For consistency, please move this to MozziGuts.h (where all the other AudioConfigXYZ.h are included).

MozziGuts.cpp
```diff
@@ -89,6 +92,12 @@ CircularBuffer <unsigned int> output_buffer2; // fixed size 256
 #endif
 
 #if IS_AVR() // not storing backups, just turning timer on and off for pause for teensy 3.*, other ARMs
+#ifdef __DAC_MCP__   // in case an external MCP4922 dac is used.
```
tfry-git: Just an idea: If you move this function to the AudioConfig(DAC).h-file, it will be easier to support additional DAC without even touching MozziGuts.cpp, in the future.

What about also supporting stereo output, though (`#if (STEREO_HACK == true)`)? From the data sheet, the MCP9422 is dual channel, so that would be an obvious use case, no? Perhaps define one output function for mono, and a separate for stereo.

Oscil.h
```diff
@@ -58,7 +58,8 @@ The usage is:
 char2mozzi.py infilename outfilename tablename samplerate
 */
 //template <unsigned int NUM_TABLE_CELLS, unsigned int UPDATE_RATE, bool DITHER_PHASE=false>
-template <uint16_t NUM_TABLE_CELLS, uint16_t UPDATE_RATE>
+template <uint16_t NUM_TABLE_CELLS, uint16_t UPDATE_RATE> //mozzi original
```
tfry-git: Nitpick: Remove the comment, so the file will remain "untouched".

mozzi_config.h
```diff
@@ -89,5 +89,16 @@ You need to have \#define STEREO_HACK true in mozzi_config.h
 */
 #define STEREO_HACK false
 
+
+/**
+  Uncomment this line if you want to use Mozzi with an external DAC of type MCP4922. Note that you will need the MCP49XX library (see AudioConfigMCP4922 for details.
+*/
+//#define __DAC_MCP__
```
tfry-git: To follow the style of existing defines, remove the __ around the name. Also, I'd suggest to name it generically ("EXTERNAL_DAC", or something). The details (exact chip) would then be configured in AudioConfigExternalDAC.h.

tfry-git:
Hi!

Ok, this is starting to look good (from my perspective). 

> moving the definition of dacMCPAudioOutput(): moving it to the header would not compile.

What you will have to do is probably to _declare_ 
```
void dacMCPAudioOutput();
void dacMCPAudioOutput2();
```

in the header, and _define_ them (e.g.) in AudioConfigExternalDAC.cpp (along with the instantiation of the DAC). Anyway, this is probably not too important _until_ support for additional DACs is added (and can still be changed, then).

One thing that I'd still like you to do is to add a few lines of documentation to MozziGuts.h (where the audio modes are detailed). Also I'll wait a week or two for others to comment (esp. since this is something I cannot test for lack of hardware). But other than that, I think this can be merged, then.

tomcombriat:
I added a few comments in MooziGuts.h for basic operations, referring to details I put in AudioConfigMCP49xx.h (especially electrical wiring).

I hope this is clear enough, let me know if you think it needs some addition!

Cheers,

Tom

PR #57, opened 2018-10-08
Revert "Fixed AutoRange::AutoRange initialization and AutoRange::next behavior."

sensorium:
Reverts sensorium/Mozzi#56
The intended behaviour is to continually adjust range_max and range_min if incoming values lie outside the ranges.  The proposed change would only have updated max and min when they exceed the initialised values, instead of allowing them to grow each time they are exceeded.

### DIFF
--- a/AutoRange.h
+++ b/AutoRange.h
@@ -24,7 +24,7 @@ public:
 	@param min_expected the minimum possible input value.
 	@param max_expected the maximum possible input value.
 	*/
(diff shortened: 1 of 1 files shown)


sensorium:
Also, the original initialisation ensures that incoming values set the growing range.  Initialising range_max to min_expected enables the first incoming value larger than min_expected to set range_max to the largest value so far encountered.  If range_max was initialised to max_expected, range_max would never change, so no auto-ranging would occur.

PR #63, opened 2019-01-02
Added STEREO_HACK support for STM32

Tinwelint78:


tfry-git:
Hi!

Sorry for the long delay. 

Please use two separate pull requests for two separate items. First part (stereo hack) looks fine to me.

Second make sense in principle, but should probably be implemented as an additional template parameter, which should allow more flexibility moving forward (e.g. floats, even larger data types, data kept in RAM, rather than flash). We had some discussion on this in https://github.com/sensorium/Mozzi/issues/47 and https://github.com/sensorium/Mozzi/issues/48, if you are interested in giving it a shot.

Tinwelint78:
Seems very reasonable, I've removed Sample16.h for now

PR #65, opened 2019-01-31
Corrected wrong behavior leading to a wrong value of the starting point of the envelop when using very small values for the release time

tomcombriat:
Hi!

This corrects the fact that, when using very small values for the release time, the envelop may start, for a new note, on a non-zero value.
I think, even though the envelop is effectively reaching 0 at the end, the last point of the line (transition) used to interpolate points between sustain value do not reach zero and its last value is used for starting the envelop.

Note: this has been for very short (<20 micro) on STM32. Did not try to reproduce on AVR but I do not see why it would behave differently.

Sincerely,

T. Combriat

### DIFF
--- a/ADSR.h
+++ b/ADSR.h
@@ -178,7 +178,7 @@ public:
 	*/
 	inline
 	void noteOn(bool reset=false){
-		if (reset) transition.set(0);
+		if (reset || !adsr_playing) transition.set(0);
 		setPhase(&attack);
 		adsr_playing = true;
 	}
--- a/ADSR.h
+++ b/ADSR.h
@@ -178,7 +178,7 @@ public:
 	*/
 	inline
 	void noteOn(bool reset=false){
-		if (reset || !adsr_playing) transition.set(0);
+		if (reset) transition.set(0);
 		setPhase(&attack);
 		adsr_playing = true;
 	}
--- a/Line.h
+++ b/Line.h
@@ -82,18 +82,15 @@ public:
 	@param num_steps how many steps to take to reach the target.
 	 */
 	inline
-	void set(T targetvalue, T num_steps)
+	  void set(T targetvalue, T num_steps)
 	{
-		T numerator;
-		numerator = targetvalue-current_value;
-		//float step = (float)numerator/num_steps;
-		T step = numerator/(num_steps ? num_steps : 1);
-		step_size= (T)step;
-		//Serial.print("numerator");Serial.print(" \t");Serial.println(numerator);
-		//Serial.print("num_steps");Serial.print(" \t");Serial.println(num_steps);
-		//Serial.print(step);Serial.print(" \t");Serial.println(step_size);
-		//step_size=(T)((((float)targetvalue-current_value)/num_steps));
-
+	  if(num_steps) {
+	    T numerator = targetvalue-current_value;
+	    step_size= numerator/num_steps;
+	  } else {
+	    step_size = 0;
+	    current_value = targetvalue;
+	  }
 	}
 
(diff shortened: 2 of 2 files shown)


tomcombriat:
Yes, the last value of the release phase is not 0 but the value for the IDLE state is correct.
As a consequence, `transition.set(0)` only has to be called once between the end of the release phase and the start of the next attack phase, that is why I put it in the 'noteOn()' function. 

Putting 'transition.set(0)' in 'update()' is also working but I was afraid that would generate many unnecessary calls to this function. I agree it would be much more readable that way.

What do you thing is the best?

tomcombriat:
Alright I will try that in the next few days.

 I think the problem effectively comes from the Line object, however, if I remember well, this would also happens when there are very few `num_steps` and not only zero: if the last point of this line is not 0 (or any wanted level) because the slope is very sharp then, and as the next phase will use the same line it will start from a wrong value.
Maybe we should ensure that the last point returned by the `Line` is effectively the target_value?

An other solution in order to maintain the fact that idle level may not be zero would be to change my line of code with `transition.set(idle.level).

tomcombriat:
Alright, I tried the modification in Line.h and it works as espected. This does not break the fact that the idle level can be non-zero, so maybe we are done?
Thanks for your help!
Tom

tfry-git:
Great! Thanks! I think this is the correct fix, then. Just to be on the safe side, I'll wait another week or so, before merging, to give Tim a chance to object. Do ping me after that, should I forget to merge!

Thomas

tomcombriat:
Hi!
Thanks for your reply!

Also, when playing a bit further with this envelope, I found out that, when putting `envelope.next()` in `updateAudio()`, which is quite necessary for sharp phases from my tests, the phases time and particularly the sustain time is limited to approximately 1.9s. This is because the step counter is coded in an `unsigned int` type. I think coding this value as an `unsigned long` would give possibilities for very long times for the envelope without a great deal for performances.

Actually, I did tests, changing `update_step_counter`, `num_update_steps` and `convertMsecToControlUpdateSteps` with `unsigned long` values and it seems to work quite well.

However, I wanted your opinion on that before comitting this new change.

Sincerely,

Tom

tfry-git:
Do not underestimate the amount of trouble that 8-bit MCUs need to go through for performing 32-bit arithmetic. And that's still a / the primary target audience. Also, your requirement of both extremely short and rather long transitions is probably not all that common.

However, a decent way out might be to add a third template parameter for the counter type. That would still default to `unsigned int`, but could be changed to a larger type by the user.

Possibly the default parameter could even depend on the hardware, but that might need more in-depth consideration.

tomcombriat:
Yes, I see your point. As I'm mainly developing for STM32 at the moment (thanks very much for your port!), I temporarily forgot that 32 bits is actually huge for an 8-bits Arduino… And that because I'm building an ADSR specific hardware for a modular synth, my needs are actually very specific…

 Sorry for bothering you with that!

Thanks for your very detailed answer!

sensorium:
Hi Tom and Thomas,

thank you both for helping to make Mozzi more useful and usable.

I'm in the process of testing Tom's suggested changes with the examples to see how they run.  If the 32 bit counter works OK with the examples, I'm open to including it in the development code.  

Thomas, your considerations also make sense to me and a template parameter would be ideal.  But I understand if Tom doesn't have time to write the extra code.

I'll get back here when I've had time to finish the tests.

Cheers
Tim

tfry-git:
Thanks Tim!

Any doubts about merging the correction to Line.h? So we can keep these two issues separate.

sensorium:
No doubts about the correction to Line.h.

PR #69, opened 2019-04-04
Manually calculate prescaler and overflow for STM32 update timer

### LINKED ISSUE #67: Strange frequency shift on STM32
I have identified a strange problem running mozzi on STM32. Playing a Sample or an Oscil is slightly faster than it should be (exactly 1.5% faster). For example, if I make a sketch that plays a sinewave at 440 Hz, on the AVR I obtain a correct 440 Hz sinewave, the same sketch running on a STM32 produces a 447Hz sound.
I verified this is independent from audio mode (HIFI or STANDARD_PLUS), and independent from the actual wavetable used.

Tinwelint78:
Fixes #67


tfry-git:
Looks good to me. As far as I can see, you have basically unpacked HardwareTimer::setPeriod(). That should get the rounding error down, indeed. I assume you have tested / scoped?

(Note: *After* this is merged, you should ```git --reset origin/master``` to get back in sync with upstream. That way your future PR's will show the relevant commits, only.)

Tinwelint78:
Yes, tested and working. Sorry for the error, I'm still fairly new to git! ;)

tfry-git:
No problem, git really isn't newbie friendly, at all.

Issue #81, opened 2019-11-02
External audio output


tomcombriat:
Hi!
As I was trying to port the MCP49XX DAC to Mozzi with an STM32 I found out that it was very hard to port different audio outputs to Mozzi as it usually has to deal with `MozziGuts.cpp`which is highly optimized and very hard to modify.

I was wondering if a good way to do this would be to allow the user to define a callback function, "extracting" the audio sample from Mozzi. I was thinking of a function like `void audioOutput(unsigned int sample)` in which the user could then send the sample to a DAC or anything else, and which would be defined in the user code, just like `updateControl` or `updateAudio` and be activated in `MozziConfig.h`.

Do you think it is worth that I try to implement such a thing (and remove the first port for an external DAC which would then be deprecated)? I think it won't affect the performances ofMozzi when using the standards outputs (PWM) and could be very useful for outputting the signal through other systems.

tfry-git:
Hi!

I think this is an excellent idea. I wonder (but am not sure) whether it would make sense to use this method, *always*. Would the compiler / linker be smart enough to optimize away the function call? If so, that would allow for an even cleaner / more consistent solution.

I suggest the existing DAC port would be kept, but moved to a dedicated include file, and implemented in terms of providing an ```audioOutput()``` function, too. If the user code includes that file, it's there, if it does not include it, it's not.

tomcombriat:
Hi again!

Alright cool! Thanks for your answer! I think this could simplify a lot the port to different devices while letting the user to define customized way of outputting the sound.

If you are okay with that, I will try in the next few days to implement that.
 In a first time I was thinking of removing the support for the external DAC from `MozziGuts.cpp` as it is not supported for other platforms than AVR (I am working on a port for STM32 which will be compatible with AVR) and implementing `void audioOutput(unsigned int sample)` and `void audioOutput2(unsigned int sample)` (for stereo). The "native" ways of outputting the sound would stay is `MozziGuts` for now, in order to give time to see if performances are remained when using these functions. 
The DAC support would come back in a second time, the time to code the new library for MCP49XX (or others) and using this new approach.

What do you think about this workflow?

tfry-git:
Sounds reasonable to me.

Some remarks:
1.  If we're making ```audioOutput()``` a part of the API, its signature should be well thought out. For one thing, it may or may not make sense to make it a 16 bit int, explicitly. (```unsigned int``` is 32 bits on ESP8266 and STM32, for instance).
[...]

PR #84, opened 2020-02-10
Added non-aliased version of saw wave (kept up to the 6th harmonic)

tomcombriat:
Playing a bit with the SAW table I found out that it was very prone to aliasing. Unlike the square it did not have a no-alias version. I made one, keeping the first 6th harmonics as a start. It is still aliased for high pitch notes but I wanted to keep the hardness of this wave.

Cheers,


aldreiperalta:
obrigado!!

tomcombriat:
Hi,
I will soon delete to start working on issue #81 .
Best,

PR #87, opened 2020-04-11
Allow to define an external audio ouput function

tomcombriat:
Hi!

This is a proposition in response to #81:
I moved outputting functions for STM32, AVR, SAMD and Teensy to a separate file. This allows the user, if `//#define EXTERNAL_AUDIO_OUTPUT true` is uncommented in mozzi_config.h, to define himself an outputting function, if wanted. This allows use of alternate ways to output the signal (DAC for instance) in a much simpler way, but retain the current behavior of Mozzi as default.

As ESP methods are way different, I let them in MozziGuts.

As discussed with @tfry-git in #81, these functions are zero centered. The BIAS is applied in AudioOutput.cpp or inside the user defined function. I also removed the previous external_dac support, which was only working with AVR. I think DACs should be used in the main sketch.

I tested (successfully) a few examples with:
- AVR (Arduino Mega)
- Teensy 3.2
- STM32
I could not test SAMD and ESP due to lack of hardware.

Best,


tfry-git:
Hi!

I'd like to look into a way to integrate ESP into this solution as well. Despite all the internal differences it would be cool, if all the architectures would "look the same" to users of the Mozzi-API. I'll try to get around to this, soonish.

Perhaps you could add an example sketch demonstrating output to an external DAC? This could also serve as a replacement for the removed existing DAC support.

tomcombriat:
Hi!
Thanks for your response. If you do not have have to look at it for ESP platforms I can give it a try, however as I do not have one I am not very confident with it.

I do not have any available DAC at the moment, but if can get my hand on one I will do an example. If not, I can make a "mocky" one, for instance with a resistance ladder, which would be interesting to show that you can output "any way".

tfry-git:
Hi!

Yes, it's growing increasingly difficult to test all supported hardware combinations. Even more reason why the modularization you are doing in this PR is so important!

As some hints (for you or for myself) on how the ESP implementation differs, and what may be needed to integrate it:

The "traditional" audio output works like this:
1. `loop()` calls `audioHook()` fills the `output_buffer` (to complicate matters, audioHook() also takes care of calling `updateControl()` every n samples).
2. A separate ISR called at `AUDIO_RATE` reads each sample from `output_buffer` and writes it to the output.

Now for ESP we actually have *two* different approaches:
A) `loop()` calls `audioHook()` writes to the hardware I2S buffer, directly. (I2S buffer gets pushed to the output, automatically, at `AUDIO_RATE`, without the need for a separate timer / ISR).
B) 1. `loop()` calls `audioHook()` writes to `output_buffer`
2. A separate timer-based ISR copies `output_buffer` to the (small) hardware Serial output buffer. *However*, the output rate is actually controlled by the Serial hardware, not by the timer-based ISR.

Both of these approaches might also be interesting for other platforms. Esp., but not only, when it comes to external DACs.

I believe the necessary bits to abstract over all the approaches will be:
1) Initializing a timer-ISR at an appropriate rate (which may or may not be `AUDIO_RATE`)
2) Your `audioOutput()` function
3) A function or macro to check whether the output buffer is full (and `audioHook()` should not generate any further samples)
4) A function or macro to write a single sample to the output buffer

Not quite sure how to do this in the most elegant way...

tomcombriat:
Hi again,

Yes, the whole idea is to get most of platform dependent stuff outside `MozziGuts`. As for ESP, no ISR was defined at `AUDIO_RATE, defining an `audioOutput()` function was not as easy. Due to lack of hardware, I do not think I will give it a try soon. 

Moreover, I personally think that Mozzi should provide ways to output sounds using the "bare hardware" as it is already the case. However, I think additional ways of outputting the sound (DAC, resistance ladder for instance) which sometimes depend on additional libraries which are not always available for every platforms should stay outside of Mozzi's core. 

Maybe, a new folder in the examples could be a good way to demonstrate the use of different hardware combinations. This could be easily updated each time a new combinations is tested.

If you are ok with, I can start this by adding an example for using a R/2R DAC. This has the advantage of not needing special hardware, to not require any additional library and to be compatible with any chip that provides digital output.

tfry-git:
Fully agreed on all of that. However, what I'd like to design into Mozzi itself is the _possibility_ to add any which output option users might think of. I think for that we'll need a few more hooks in addition to your `audioOutput()`. Importantly an option to replace Mozzis default output buffer with a custom mechanism. I am not sure whether I'm correct on the details, though.

tomcombriat:
Hi!
I added an example using an R/2R DAC. As this is using `digitalWrite()`, which is quite slow, it can output only 6bits without troubles. A lot of optimizations can be done, using registers for instance, but that is platform dependent and wanted to keep it simple!

tfry-git:
Ok, thinking about it, let's skip most of my comments for now. I think I now have an idea how to realize those points, and it's mostly orthogonal to your work, so no reason for a hold-up.

Now, looking at this PR, I think it's looking good, but I do have one remaining concern: In most cases this will involve one additional function call per audio frame, (`audioOutput()`). I believe, the compiler will _probably_ not be able to inline this call. This might be a problem for our slowest architecture (AVR), and the sketches that are close to the limit, already.

1. Have you done testing on this aspect? I.e. do the more computation heavy examples still work on AVR?
2. Although a bit messy, perhaps it would be possible to define/declare `audioOutput()` as a function only if `EXTERNAL_AUDIO_OUTPUT` is defined, and as a macro otherwise? That would probably require AudioOutput.cpp to be `#include`ed in MuzziGuts.cpp.

tomcombriat:
Hi!
I was actually wondering at the beginning if it would be worth to `inline audioOutput()`. In that case, and if the compiler is actually inlining it, I though that would be completely equivalent to the way Mozzi was working originally. But maybe there are some underlying stuff that I do not know/understand. I was also wondering if passing the audio samples to this function using references instead of values could not speed up the thing.
I did some testing, usually with the FM synth example from Mozzi, but not extensively. I do not have a scope so I will try to make to compare the performances using a sketch which is *on the edge* already. I will test the different combinations and let you know.

I am also working on an example using a 74HC595 shift register which I think could be useful. If it does work well I will add it in a few days.

tfry-git:
Ok, cool.

My understanding (which may be wrong) is that the compiler will not be able to inline the function, because the definition is in a separate compilation unit (AudioOutput.cpp being compiled, separately). Declaring it inline _and_ bringing it into the header (AudioOutput.h) _might_ work, though.

tomcombriat:
Hi!
I did some tests. 
### Method:
I used a basic Sinewave2048 but with an increased number of oscillators. Only one is actually outputting sound but all being refreshed in `updateAudio()`. The score I present here is the highest possible number of oscillators without the circular buffer getting empty, so that the signal is clean. The highest the score the better. The `CONTROL_RATE` is 64 and this is done with an Arduino Mega.

### Results:
- original Mozzi (without `audioOutput` implemented): 12
- default `audioOutput()`: 12
- default `audioOutput()` with inlining: 12
- user defined `analogWrite()`: 9
- using a shift register as presented is the last example: 9

Without any surprises, the default Mozzi is performing better than an `analogWrite` as I uses registers, however it does not seem to do better than the solution presented in this PR. Inlining does not seem to have an effect either, but I did not put the function in the same compilation block. Anyway, in no case it could perform better than the original version. My last example is not performing to bad even if not as good as Mozzi's default, the wave seems a bit cleaner.

Best,

tomcombriat:
Hi,
In the end I moved everything in `AudioOutput.h`. This way we are sure performances are the same than before when using the default output functions.
I will receive some DACs in the future and will provide other examples of their integration in Mozzi by `audioOutput` functions, but maybe that could come in another PR as I am not sure to receive them soon.

Best,

tfry-git:
Yes, I think that is most in line with the Mozzi philosophy of optimizing as much as possible (at least for AVR). From my side this is ready to go in. @sensorium : any objections?

tomcombriat:
(I just modified the example, where left and right were the opposite of what I did in the main code, this is only naming, but no need to confuse someone with that)

tomcombriat:
Hi,
Any news from @sensorium ? I think I will start working on examples for MCP4XXX DACs soon, but I think it will be better placed in another PR. I do not know how long it will take me.

Best,

tfry-git:
As you can see, I've merged your PR. Seeing that it neither needs more CPU, nor more programming - unless and until you want to use the new feature - there should not be any problem, I think.

Independently of this, I'd like to encourage you to read up on [git feature branches](https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow). It may look a bit scary at first, but it's designed specifically for working on independent features in parallel. I'm sure you love the concept, once you get the hang of it!

sensorium:
Please feel welcome to go ahead in any directions that interest you, Thomas and Tom.  I don't have the hardware so far to test what you're doing, but am following along with the conversation.

dietervandoren:
Hi gents, as some of you have noticed in the google group discussions I've been toying with ESP32 and external audio output, until Tom explained to me that ESP platforms are not included in the external audio output structure. I now have modded the current Mozzi master branch to work with ESP32 and PT8211 DAC, though outside of the external audio output structure. I see talk in this thread of plans to expand external audio output to include ESP, but I don't see steps taken in that direction. I would like to contribute my code for ESP32+PT8211 but I am not capable of coding the extra steps laid out by Thomas to make it fit in the external audio output structure. Does it help if I make a pull request with my current code (tested working), so that more skilled coders can turn it into a proper external audio output structure? Or perhaps my ESP32+PT8211 code is of interest without plans to move it to external audio output? 

Greetings, Dieter

dietervandoren:
Oh and can do the same for ESP32 with internal DAC. It's just a matter of changing some I2S settings.

But it'd be great to have Mozzi experts review both because I'm not 100% sure the signal is passed correctly to the I2S interfaces. Timing might be off.

tomcombriat:
Hi! Just a quick comment (I am not a maintainer of Mozzi).
Sounds great! I am *really* not the expert in ESP/I2S, and do not have the hardware to test… I think that I would be great to have a coherent layout for all platforms and this could help doing so!
Hopefully @tfry-git who is the expert might give you some clues!

Best,

dietervandoren:
@tfry-git do you have thoughts on this?

Alternative is that I make a pull request with my ESP32 internal DAC and PT8211 methods incorporated in MozziGuts and selector defines in MozziConfig.

tfry-git:
Hi!

Sorry, not much time to look into this, right now. I do suggest to create a pull request with your work, indeed, then we can see what should still be changed. If you can, try to keep your code similar to the existing code e.g. for the ESP8266 (I assume you'll also be using a DMA output queue and thus no separate timer). If we are going to change some of that in the mid term, it will be easier to adjust you code at the same time, then.

Regarding exact timing, I suggest not to worry *too* much, at first. Try to test as well as you can. However, as long as existing architectures keep working, your port does not have to be perfect in the first attempt. (As a matter of fact, in my STM32-port, it turned out that the pitch was a bit off, initially, too).

dietervandoren:
Actually the solution does use a separate timer. I pulled it from this thread and fixed the apparent issues: https://groups.google.com/g/mozzi-users/c/qw2nIkcVaZE/m/hjJef464CQAJ

I'll get on preparing the pull request now.

Thanks, Dieter

dietervandoren:
It's here: https://github.com/sensorium/Mozzi/pull/97

First time github upload so please tell me if I didn't do things the best way. (Ok I did it once before but that's like 10 years ago...)

Anyeos:
Hello I am interested on a user defined audioOutput because I need time critical handling routines inside an ISR. When I implement my own ISR Mozzi does not work anymore (because the callback function of ISR is now other).
So using my own audioOutput function can help of solving that.
Can we have some form of defining the polltime too or reading the value in a constant or var?

PR #90, opened 2020-05-24
Added examples using DAC MCP49XX/48XX with Mozzi

tomcombriat:
Hi,

This provides examples for using SPI DAC MCP49XX or MCP48XX with Mozzi, using the `audioOutput` function.
Three examples are provided:
- FMsynth_MCP4921_mono_12bits is a simple mono 12 bits based on the FM example,
- Stereo_Pan_MCP4922_stereo_12bits is a simple stereo 12 bits, using Stereo_Hack,
- MCP4922_mono_24bits is a simple mono 24 bits, using the same technique than the HiFi mode of Mozzi.

#### A few notes:
- I had to modify the library I was using to provide support for non-AVR platform. The modified library is available [here](https://github.com/tomcombriat/DAC_MCP49XX) (this is written in the sketches)
- All examples work with STM32 board
- First example works fine on Arduino Mega, but the second is not so smooth. I think there is a compatibility problem between the Mega and the PortWrite technique used in the library. I would like to test on an Uno, but do not have one now. This will likely be fixed by updating the library in the future.
- Last example is impossible to get working on AVR because the audio sample does not fit the 16 bits `int` container. However, 16 bits (2*8bits) could be achieved. This works well on STM32, even if it is a bit overkill…

Let me know what you think!
Best,

Tom

### DIFF
--- a/examples/13.External_Audio_Output/FMsynth_MCP4921_mono_12bits/FMsynth_MCP4921_mono_12bits.ino
+++ b/examples/13.External_Audio_Output/FMsynth_MCP4921_mono_12bits/FMsynth_MCP4921_mono_12bits.ino
@@ -10,7 +10,7 @@
     MCP4921   //  Connect to:
     -------       -----------
     Vdd           V+
-    CS            any digital pin defined by SS_PIN (see after), or pin 10 on UNO / 11 on Mega if you are using Portwrite
+    CS            any digital pin defined by SS_PIN (see after), or pin 7 on UNO / 38 on Mega if you are using Portwrite
     SCK           SCK of Arduino
     SDI           MOSI of Arduino
     VoutA         to headphones/loudspeaker
@@ -60,7 +60,7 @@ Smooth <int> kSmoothNote(0.95f);
 
 
 // External audio output parameters and DAC declaration
-#define SS_PIN -1  // if you are on AVR and using PortWrite, you do not care what you put here.
+#define SS_PIN 38  // if you are on AVR and using PortWrite you need still need to put the pin you are actually using: 7 on Uno, 38 on Mega
 #define AUDIO_BIAS 2048  // we are at 12 bits, so we have to bias the signal of 2^(12-1) = 2048
 DAC_MCP49xx dac(DAC_MCP49xx::MCP4921, SS_PIN);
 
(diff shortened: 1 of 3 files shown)


sensorium:
Hi Tom,
I'm happy to merge this - let's see if Thomas has any thoughts about it.

tfry-git:
Hi,

I have not tested any of this, but it's looking good to me. I'm sure, these examples will be useful to many! Two general comments:

1. Admittedly the upstream DAC_MCP49XX seems inactive, but for good measure, you should submit a PR with your changes, there, too.

2. It would really be good to get some testing on Arduino Uno (or other 328P-based board), since this is still "the" Arduino. IMO it's ok, if some examples don't work there, but it would be a good idea to have that fact clearly documented (even a bit clearer / more upfront than you did in your third example, to be precise). Not a show-stopper for this PR, though.

Regards
Thomas

tomcombriat:
Hi Tim and Thomas,
Thanks for your feedback! Just to comment:
- I already removed the upstream DAC support in PR #87. I am far more satisfied with this version which is much more flexible.
- I will test on a Uno soon. I think, the DAC library is in cause as unfortunately portwrite is using one of the SPI pin. I will probably come with a fix for Mega soon on this side. This should not have any effect on Mozzi, except maybe for the pin names in the comments at the top of these examples.
- I can make a clearer warning for the last example if you want. Such technique should work to output 16 bits on an Arduino.
- the sound is very nice without any filtering from what I have tested! 

Best,
Tom

tomcombriat:
Hi!

I added a commit:
- I modified the DAC library, so that, when using portwrite to speed up on AVR, this does not conflict with SPI on Arduino Mega. As a consequence, the pin used for portwrite is not the same anymore so I updated the comments in the examples for Mozzi as a consequence.
- Stereo_Pan_MCP4922_stereo_12bits now tested and working for both Mega and UNO (I found one). Other boards should also work fine using digital pin 7 for SPI CS.
- I emphasized a bit the fact that no more than 16bits can be achieved on AVR boards in the 24bits.  

Best,
Tom

tomcombriat:
Hi!
I saw on the forum that people start to play with these DACs again. We will probably have good feedback how good this is working! I do not have a lot of time these days but I will soon make a proto board, with good and stable electronic to evaluate properly the performances.
Best,

PR #95, opened 2020-09-30
Added a third template parameter to ADSR to allow long envelopes at AUDIO_RATE

tomcombriat:
Hi!

This adds a third template parameter to the ADSR, affecting the update steps.

It allows to create long envelopes at AUDIO_RATE as the `unsigned int` initially used can be replaced by `unsigned long` for instance; the initial implementation does not allow ADSR longer than 3.6s (if my memory is correct) at AUDIO_RATE. This reduces performances on AVR, but does not affect native 32bits platforms like STM32 or ESP.

A default parameter (`unsigned int`) is specified thus, this does not change the behavior of existing sketches.

An example, basically copy pasted from an existing one, with small adaptations is also added.

Let me know if you have any comments!

Best,

### DIFF
--- a/ADSR.h
+++ b/ADSR.h
@@ -41,22 +41,22 @@ Template objects are messy when you try to use pointers to them,
 you have to include the whole template in the pointer handling.
 */
 
-template <unsigned int CONTROL_UPDATE_RATE, unsigned int LERP_RATE>
+template <unsigned int CONTROL_UPDATE_RATE, unsigned int LERP_RATE, typename T = unsigned int>
 class ADSR
 {
 private:
 
 	const unsigned int LERPS_PER_CONTROL;
 
-	unsigned int update_step_counter;
-	unsigned int num_update_steps;
+	T update_step_counter;
+	T num_update_steps;
 
 	enum {ATTACK,DECAY,SUSTAIN,RELEASE,IDLE};
 
 
 	struct phase{
 		byte phase_type;
-		unsigned int update_steps;
+		T update_steps;
 		long lerp_steps; // signed, to match params to transition (line) type Q15n16, below
 		Q8n0 level;
 	}attack,decay,sustain,release,idle;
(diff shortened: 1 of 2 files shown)


tfry-git:
I have not tried this, but it makes a lot of sense to me.

One detail to consider might be that ```unsigned int```, in contrast to ```uint_32t``` is not defined the same way on all platforms (16 bit on AVR vs. already 32 bit on the 32 bitters). This may actually be quite beneficial, as the 32-bitters tend to have both much more RAM and fast 32-bit arithmetic, and so a different default is probably even desirable. But just to point it out.

tomcombriat:
Thanks for your comment!

Yes, actually, this changes this behavior because of 
`T convertMsecToControlUpdateSteps(unsigned int msec){
		return (T) (((uint32_t)msec*CONTROL_UPDATE_RATE)>>10); // approximate /1000 with shift
	}`

which had an explicit `(uint16_t)` cast, thus reducing *de facto* the possible duration of envelopes on 32bits fellows. Now, it will be casted as 32 bits on 32 platforms and probably the third argument of the template won't be needed (or only for extremely long envelopes). AVR will not see any differences…

This is really a minor change, but can help having precise envelopes with long sustain times, which is usually the expected behavior for key-played synthesizers.

Best,

tfry-git:
Yes. To be clear: I'm all for it.

Will merge in a week or so, if there are no objections (and if @sensorium does not beat me to it... and if I don't forget - ping me, if I do).

sensorium:
I'm for it too.  Does it need any more testing?  I can only do avr's but happy to test on one if needed.

tomcombriat:
I tested that on AVR already (Arduino Mega), but a double check can never hurt. I haven't check on STM32, but this is basically a modification I was applying every-time I would use Mozzi on STM32 so I am very confident.

Best,

CZHO404:
this would be great addition to mozzi

tomcombriat:
Just tested with STM32: works well, even without explicitly defining the third parameter as `unsigned int` are wider on STM32 than on AVR. Works with shorter type too (`short` on STM32) but dunno who would need that…

PR #97, opened 2020-10-30
ESP32 support (internal and PT8211 DAC)

dietervandoren:
Optional choices for PT8211 DAC and internal DAC in mozzi_config.h, though internal DAC is buggy. 

I have a working solution for internal DAC based on the Mozzi version from around january 2020, because that's what zcarlos's solution in the google forum was built on. I fixed that to work better. However when I port those I2S settings to the current version of Mozzi I get a chopped up signal that I can't get into shape by bitshifting ranges, which helped before.  

PT8211 works, though I am unsure whether timing of sample calculation and output to the DAC are properly synced.

I checked whether an Arduino Duemilanove still works. Seems fine.

### DIFF
--- a/MozziGuts.cpp
+++ b/MozziGuts.cpp
@@ -37,7 +37,6 @@
 #elif IS_ESP8266()
 #include <Ticker.h>
 #include <uart.h>
-#elif IS_ESP32()
 #endif
 
 
@@ -101,7 +100,7 @@ CircularBuffer<unsigned int> output_buffer2; // fixed size 256
 	#include "freertos/queue.h"	
 	hw_timer_t * timer = NULL;
 	int i2s_num = 0;
-#if (ESP32_PT8211 == true)	
+#if (ESP32_AUDIO_OUT_MODE == PT8211_DAC)	
 	i2s_config_t i2s_config = {
 		.mode = (i2s_mode_t)(I2S_MODE_MASTER | I2S_MODE_TX),
 		.sample_rate = AUDIO_RATE,
@@ -119,7 +118,7 @@ CircularBuffer<unsigned int> output_buffer2; // fixed size 256
 	  .data_out_num = 33, 
 	  .data_in_num = -1 
 	};
-#elif (ESP32_internal == true)
+#elif (ESP32_AUDIO_OUT_MODE == INTERNAL_DAC)
 	i2s_config_t i2s_config = {
 		.mode = (i2s_mode_t)(I2S_MODE_MASTER | I2S_MODE_TX | I2S_MODE_DAC_BUILT_IN),
 		.sample_rate = AUDIO_RATE,
@@ -500,9 +499,9 @@ void samd21AudioOutput() {
 
 #elif IS_ESP32()
 static void IRAM_ATTR esp32AudioOutput() {
-#if (ESP32_PT8211 == true)	
+#if (ESP32_AUDIO_OUT_MODE == PT8211_DAC)	
 	uint32_t outputEsp32=(uint32_t)output_buffer.read();
-#elif (ESP32_internal == true)
+#elif (ESP32_AUDIO_OUT_MODE == INTERNAL_DAC)
 	uint32_t outputEsp32=(uint32_t)output_buffer.read()<<8;
 #endif
 	i2s_write_bytes((i2s_port_t)i2s_num, (const char *)&outputEsp32, sizeof(uint32_t), 0);
@@ -559,14 +558,14 @@ static void startAudioStandard() {
 #elif IS_ESP32()
   #ifdef DO_NOT_INIT_ESP32_I2S
   #else
-	#if (ESP32_PT8211 == true)
+	#if (ESP32_AUDIO_OUT_MODE == PT8211_DAC)
 	i2s_driver_install((i2s_port_t)i2s_num, &i2s_config, 0, NULL);
 	i2s_set_pin((i2s_port_t)i2s_num, &pin_config);
 	//i2s_set_sample_rates((i2s_port_t)i2s_num, 44100);
 	i2s_set_sample_rates((i2s_port_t)i2s_num, AUDIO_RATE);    	
 	i2s_zero_dma_buffer((i2s_port_t)i2s_num);
 	
-	#elif (ESP32_internal == true)
+	#elif (ESP32_AUDIO_OUT_MODE == INTERNAL_DAC)
 	i2s_driver_install((i2s_port_t)i2s_num, &i2s_config, 0, NULL); 
 	i2s_set_pin((i2s_port_t)i2s_num, NULL);
 	i2s_set_sample_rates((i2s_port_t)i2s_num, AUDIO_RATE);
--- a/MozziGuts.cpp
+++ b/MozziGuts.cpp
@@ -37,6 +37,7 @@
 #elif IS_ESP8266()
 #include <Ticker.h>
 #include <uart.h>
+#elif IS_ESP32()
 #endif
 
 
@@ -91,6 +92,46 @@ CircularBuffer<unsigned int> output_buffer;  // fixed size 256
 CircularBuffer<unsigned int> output_buffer2; // fixed size 256
 #endif
 //-----------------------------------------------------------------------------------------------------------------
+#endif
+
+// ESP32
+#if IS_ESP32()
+
+	#include <driver/i2s.h>
+	#include "freertos/queue.h"	
+	hw_timer_t * timer = NULL;
+	int i2s_num = 0;
+#if (ESP32_PT8211 == true)	
+	i2s_config_t i2s_config = {
+		.mode = (i2s_mode_t)(I2S_MODE_MASTER | I2S_MODE_TX),
+		.sample_rate = AUDIO_RATE,
+		.bits_per_sample = I2S_BITS_PER_SAMPLE_16BIT,
+		.channel_format = I2S_CHANNEL_FMT_RIGHT_LEFT,
+		.communication_format = (i2s_comm_format_t)(I2S_COMM_FORMAT_I2S | I2S_COMM_FORMAT_I2S_LSB),
+		.intr_alloc_flags = ESP_INTR_FLAG_LEVEL1, 
+		.dma_buf_count = 8,
+		.dma_buf_len = 8, 
+        .use_apll = 0 // Use audio PLL
+	};
+	i2s_pin_config_t pin_config = {
+	  .bck_io_num = 26, 
+	  .ws_io_num = 25, 
+	  .data_out_num = 33, 
+	  .data_in_num = -1 
+	};
+#elif (ESP32_internal == true)
+	i2s_config_t i2s_config = {
+		.mode = (i2s_mode_t)(I2S_MODE_MASTER | I2S_MODE_TX | I2S_MODE_DAC_BUILT_IN),
+		.sample_rate = AUDIO_RATE,
+		.bits_per_sample = I2S_BITS_PER_SAMPLE_16BIT,
+		.channel_format = I2S_CHANNEL_FMT_RIGHT_LEFT,
+		.communication_format = (i2s_comm_format_t)(I2S_COMM_FORMAT_I2S_MSB),
+		.intr_alloc_flags = ESP_INTR_FLAG_LEVEL1, 
+		.dma_buf_count = 8,
+		.dma_buf_len = 8,
+        .use_apll = 0 
+	};
+#endif
 #endif
 
 #if IS_AVR() // not storing backups, just turning timer on and off for pause for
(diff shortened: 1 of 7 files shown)


mozzi_config.h
```diff
@@ -91,5 +91,11 @@ You need to have \#define STEREO_HACK true in mozzi_config.h
 
 //#define EXTERNAL_AUDIO_OUTPUT true
 
+// ESP32 with PT8211 DAC, requires AUDIO_RATE 32768
+#define ESP32_PT8211 true
```
tfry-git, line 95: Since this is specific to ESP32, please move this to AudioConfigESP32.h.

Also (nitpicking) consider using defines along the way laid out there (for ESP_AUDIO_OUT_MODE). Keep in mind that more than just these two options (internal DAC vs PT2811) may exist in the future. A true/false define does not scale well with that.

MozziGuts.cpp
```diff
@@ -503,6 +555,33 @@ static void startAudioStandard() {
...
+	i2s_set_pin((i2s_port_t)i2s_num, NULL);
+	i2s_set_sample_rates((i2s_port_t)i2s_num, AUDIO_RATE);
+	i2s_zero_dma_buffer((i2s_port_t)i2s_num);
+	i2s_set_dac_mode(I2S_DAC_CHANNEL_RIGHT_EN);
+	#endif
+	
+	timer = timerBegin(0, 80, true);
+	timerAttachInterrupt(timer, &esp32AudioOutput, true);
+		// zc - 30 microseconds is ok for now
+	timerAlarmWrite(timer, 30, true);
```
tfry-git, line 580: Are you sure, the timer needs to run this fast? On the 8266, I2S already keeps a buffer, I'd be surprised, if it didn't on the ESP32. You can probably write a bunch of samples at once in each pass.

Also, perhaps I can trick you into looking into the I2S output on the 8266? This bypasses the output_buffer, completely, using only the I2S buffer, and I2S output rate, instead. (ESP_AUDIO_OUT_MODE == EXTERNAL_DAC_VIA_I2S)

tfry-git (commented):
Hi,

here's just a few nitpicks for now, hopefully not hard to fix. I haven't really looked into the i2s setup, or thought about why it may not work for the internal dac, so far.

Thomas

mozzi_config.h
```diff
@@ -58,8 +58,8 @@ and comment out \#define AUDIO_MODE STANDARD and \#define AUDIO_MODE STANDARD_PL
 http://blog.makezine.com/2008/05/29/makeit-protodac-shield-fo/ .
 Mozzi-users list has a thread on this.
 */
-#define AUDIO_RATE 16384
-//#define AUDIO_RATE 32768
+//#define AUDIO_RATE 16384
+#define AUDIO_RATE 32768
```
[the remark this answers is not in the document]
dietervandoren: Forgot to restore after testing...

mozzi_config.h
dietervandoren: Gotcha.

MozziGuts.cpp
dietervandoren: This is the part where it gets real difficult for me. I certainly tried to analyze and replicate the ESP8266 methods but didn't get far with that. For example, i2s_write_lr() doesn't seem to exist for ESP32. So there's a good deal of engineering to be done and it's more than I can handle. The solution I'm using was given by zcarlos in the mozzi google group and I think he pulled it form a web resource elsewhere that I came across too.  

So for the timer interval, I tried different values and nothing gave a better sounding result (mostly garble really) but I don't know in what ballpark range I should be looking. Sound quality with this value (and combination of I2S settings) isn't all that great though.

dietervandoren:
@tfry-git I processed your comments. Thanks for looking into this!

tfry-git (commented):
Thanks. There are still a few changes left over in mozzi_config.h, which are no longer needed.

Regarding the DMA output buffering, I think I now see that the real problem is the absence of a function to tell whether or not the DMA buffer is full. This makes it hard to write a direct equivalent of the ESP8266 code. At the same time, my (mozzi-)gut feeling is that inefficient buffering may be the cause of the current output problems. (*)

Maybe the way forward will be to merge your changes, but to document the ESP32 port as still unstable and subject to changes.

How well does output over a PT8211 work, BTW?

--

(*) Note: i2_write_bytes() allows to time out, if no buffer space is available, and this condition would be visible from the bytes_written parameter. Something along the following lines might work:

```
inline void audioHook() {
   static bool have_prev_sample = false;
   if (!have_prev_sample) {
      static uint32_t prev_sample = updateAudio();
      if (!update_control_counter) {
        update_control_counter = update_control_timeout;
        updateControlWithAutoADC();
      } else {
        --update_control_counter;
      }
   }
   size_t bytes_written;
   i2s_write(i2s_num, &prev_sample, sizeof(unit32_t), &bytes_written, 1);
   have_prev_sample = (bytes_written == 0);
}
```

That's still a bunch of overhead, but the ESP32 should be able to handle it. Also *(in theory)*, this should allow to fill the DMA buffer, completely, while making sure not to block when the buffer is full. No separate timer would be needed then. Feel like giving this a try?

dietervandoren:
I tried that code but the compiler complains:

C:\Users\DTR\Documents\Sync_2\Arduino\libraries\Mozzi-master-ESP32-DAC-PT8211-v2\MozziGuts.cpp: In function 'void audioHook()':
C:\Users\DTR\Documents\Sync_2\Arduino\libraries\Mozzi-master-ESP32-DAC-PT8211-v2\MozziGuts.cpp:466:24: error: 'prev_sample' was not declared in this scope
    i2s_write(i2s_num, &prev_sample, sizeof(uint32_t), &bytes_written, 1);
                                 ^

The declaration of prev_sample is indeed conditional. Also, have_prev_sample is declared as false every time again, so  if (!have_prev_sample) will always return true. When I move the 2 declarations out of it I get an error I don't know what to make of:

C:\Users\DTR\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.3/tools/sdk/include/driver/driver/i2s.h:298:11: note:   initializing argument 1 of 'esp_err_t i2s_write(i2s_port_t, const void*, size_t, size_t*, TickType_t)'
 esp_err_t i2s_write(i2s_port_t i2s_num, const void *src, size_t size, size_t *bytes_written, TickType_t ticks_to_wait);
               ^


Re: PT8211 audio quality. It's not terrible but it's got a lot of small crackles. I'm gonna do a more precise comparison of AVR PWM vs ESP32 PT8211 and let you know.

dietervandoren:
Also, are you in Germany? I'm happy to send you a Lolin D32 and PT8211 if that helps getting things going.

tfry-git:
You are right, I messed up the scoping of prev_sample. No problem for initialization of have_prev_sample, since it is declared static (which makes it behave like a global variable, but only "visible", locally). However have_prev_sample is actually redundant, too. Try this version:

```
inline void audioHook() {
   static uint32_t prev_sample = AUDIO_BIAS; // intialisation only done once
   size_t bytes_written;
   i2s_write(i2s_num, &prev_sample, sizeof(unit32_t), &bytes_written, 1);
   if (bytes_written != 0 {
      prev_sample = updateAudio();
      if (!update_control_counter) {
        update_control_counter = update_control_timeout;
        updateControlWithAutoADC();
      } else {
        --update_control_counter;
      }
   }
}
```

Yes, I am located in Germany, and having an ESP32 to spare could indeed be helpful in allowing me to test (I do sort of "have" one, but it's built into my kid's MP3-player - non-Mozzi-based).

dietervandoren:
Can you email me your address? I'll mail you the components: dtr.vndrn@gmail.com

I have yet to test your updated code.

dietervandoren:
I fixed a couple of small thingies and now this code compiles, but sound is very crunchy. Sine sounds much more like a square. See attached scope readout of what should be a 1500Hz sine. 

`  static uint32_t prev_sample = AUDIO_BIAS; // intialisation only done once
   size_t bytes_written;
   i2s_write((i2s_port_t)i2s_num, &prev_sample, sizeof(uint32_t), &bytes_written, 1);
   if (bytes_written != 0) {
      prev_sample = updateAudio();
      if (!update_control_counter) {
        update_control_counter = update_control_timeout;
        updateControlWithAutoADC();
      } else {
        --update_control_counter;
      }
   }`

sine_1500

tfry-git:
Hm, that really does not look too good, yet. With a bit of luck, I'll have your ESP32 by the weekend, though, and try some things.

Am I looking at your scope readout right, though? Isn't that a 3kHz (scrambled) sine, when you were trying to produce a 1.5kHz sine? Any chance the timer-based output (esp32AudioOutput()) is still active in addition to this?

dietervandoren:
Good call. It does it indeed look like half the period one would expect at 1500Hz. Will have to double check on the scope freq analyzer. 

I'm positive about the timer construct being disabled. Though I now realized I was running it at AUDIO_RATE 16384 instead of 32768. Not sure if that can cause freq doubling.

It will be a couple of days before I can get back on it.

tfry-git:
Your package arrived. Thanks! I can now hear the problem...

The key bug in my code is the lack of proper shifting. With ```prev_sample = (updateAudio() + AUDIO_BIAS) << 8;``` I can get the internal dac output to work (well, mostly; I'm still experimenting with the various parameters).

So the good news is I think we're inching closer to a solution, but I'll need to fiddle around some more (and esp., I have not even tried the external DAC, yet). I'll keep you posted.

dietervandoren:
Great!

tfry-git:
BTW, I've pulled your changes into a branch in my own fork, so I have a place to experiment (while keeping the intermediary results): https://github.com/tfry-git/Mozzi/tree/pr/esp32 . When done, I guess I'll have to create a pull request against your fork, so we can get everything in here, neatly, but I guess that should be one of the easier steps...

dietervandoren:
Yes I see. Anything I can do? Test some things?

tfry-git:
Yes, it would be nice, if you could do some testing. 

Next steps: If things work for you, you can merge my PR into your repo. Then it should become part of this PR, automatically, and I can merge everything into Mozzi. After that I'd look into way to make a higher AUDIO_RATE the default for ESP32/8266 and others, but not AVR.

dietervandoren:
Both internal DAC and PT8211 work. Cool! 

I did some quick first comparison between my hacked solution and your new one. On long fade out tails I'm noticing that yours is considerably more noisy and aliased. A reason for that I could see is that in my lib, at the output stage (esp32AudioOutput() ), there is not a <<8. There is in your audioHook().

In the particular sketch in updateAudio() I have this: (this sketch has a total number of 36 osc's though they never sound all together)

`  long mix =
    (drone_A >> 2) * mute_A + (pulses_L_A ) +
    (drone_B >> 2) * mute_B + (pulses_L_B ) +
    (drone_C >> 2) * mute_C + (pulses_L_C ) ; // set mix levels by bitshifting (= fast division by powers of 2)
  return (int)(mix >> 2);`

To make it work with your solution I have to change it to (mix>>10). I think that's where a lot of resolution gets lost. How can we get the most of the available bit depth?

tfry-git:
Ah right. Shows how little testing I did so far. Should now be fixed in my PR.

(As a side note, I'm also working on a solution to take care of appropriate shifting, so that sketches written for an 8 bit output mode will not have to be adjusted, manually, when switching to a 16 bit output mode, and vice versa; but that's for a different PR, too.)

tfry-git:
Hi! Did you get around to do another round of testing? Are you going to, short term?

Otherwise I'd suggest to just merge. Like I said, I think a new port does not have to be perfect, right away. For all I am aware, this one is working reasonably well. So let's no worry too much before merging.

(BTW, I also plan on adding a PDM output mode, but before that, I believe MozziGuts.cpp is in rather desperate need of a cleanup: https://github.com/sensorium/Mozzi/pull/98).

dietervandoren:
Hey, I've been in bad shape last week and didn't get around to continue on it. I have the next session planned on thursday. If you 'd like to go ahead already though, please do.

tfry-git:
Sorry to hear that. Thursday's soon enough, by all means. Essentially I had simply started wondering whether you were waiting for me for the next step.

dietervandoren:
Something's not quite right. Internal DAC seems OK. Noisy, but I guess that's to be expected. Right?

With PT8211 something's off. With the sinewave example, no matter what bitshifting I do before return, I never get a full continuous sine on the scope. It's wrapping around the wave. This is with <<8:

sine wrapped

tfry-git:
I could have sworn that is not what I saw, when I committed... Anyway, next attempt uploaded.

> Internal DAC seems OK. Noisy, but I guess that's to be expected. Right?

I guess so. I'll try to add a PDM output mode, as an alternative to the internal DAC. Hopefully that will allow for a bit better quality even without an external DAC.

dietervandoren:
This works and is quite a bit cleaner sounding than my hack. Hooray!

tfry-git:
Merged. Thanks!

Any comments on https://github.com/sensorium/Mozzi/pull/99 ?

Issue #72, opened 2019-04-12
Dual-license with GPL?


mangtronix:
Is it possible to make Mozzi available under GPL license? That would make it possible to combine with GPL licensed code. The CC-BY-NC-SA is not compatible with GPL.

Creative Commons itself "[recommends and uses free and open source software licenses for software.](https://creativecommons.org/about/program-areas/software/)"

I want to publish ESP32 support for Mozzi using the ESP8266Audio library, which is GPL licensed. Mozzi could add the GPL as an option in addition to the existing license.

sensorium:
My understanding is that if your application is non-commercial you can share Mozzi under CC-BY-NC-SA.  What is the hurdle to combining with GPL code?  

If you have a commercial application, we can negotiate a license.

tfry-git:
I believe the problem is the following:
- If you create a work using ESP8266Audio, then you *must* license that under GPL.
- If you are create a work using Mozzi, then you *must* (in be absence of another agreement) license that under CC-BY-NC-SA.
- Although the two are similar in spirit, they are not compatible, thus you have no option to create a work using *both* ESP8266Audio *and* Mozzi.

One way out would be, if Mozzi would *optionally* allow you to distribute your derived work under GPL. I'd be sympathetic to that idea, but clearly it's your call, alone, to make.

tfry-git:
I will add that the GPL does not preclude commercial use. All it says is that you must offer your source code under the GPL.

DCooper-nz:
Would be great if this could be worked out as esp32 support would be fantastic.

wyan:
What's the reason for the ESP8266 port not requiring ESP8266Audio but the ESP32 requiring it?

tfry-git:
That comment was before support for the ESP32 was added to Mozzi (very recently). Neither port requires ESP8266Audio. Only the code the OP seemed to have in mind, did. The problem about a license clash if using both libs in the same project still stands.

sensorium:
Hi, it's taking a while but I've been looking at licenses and I don't really understand how a dual license works, or how to present it with a project.  Is there an example somewhere?  Thanks...

wyan:
[...]

PR #98, opened 2020-11-18
Further abstraction of output handling (on top of https://github.com/…

tfry-git:
…sensorium/Mozzi/pull/87) to allow:

- sane handling of stereo (and potentially further channel formats)
- integrating the ESPs into the audioOutput() framework
- removing a lot of redundant code, and
- making the platform ifdefs more readable, by slicing them out of the main logic as much as possible
- "automatic shifting" for sketches that use the appropriate functions.

This is still WIP, and entirely untested.


tfry-git:
An update on this PR (I've now changed it from a "draft" to a real request):

I've started some testing, with good results, but of course this is a rather largish diff, and nowadays there are a whole lot of combinations to test. I'll try to do some more systematic testing, and share the results.

At any rater, however, I think, this is a very important update to keep the code-base maintainable in the long term, as it separates most of the platform-specifics from the common logic. Compare this PRs variant of audioHook() with that in master to get an idea of what I mean. Further simplifications may be possible (e.g. WRT setting up the required timers), but this addresses a sizeable chunk of the #ifdef-mess.

For all I care, this PR is not urgent. However, since it does affect a whole lot of code, it would make sense to (hopefully) merge this, before the next major change to MozziGuts.cpp. So if you have anything in mind along those lines, please take some time to look at this, first.

tfry-git:
Alright, I now adjusted all the existing examples to the "new best practice" after this PR. That means, examples will now run unmodified on all boards (hooray!). The separate HIFI mode examples are also pretty much obsolete, due to this, but I left them for now.

I tested many, but not all, so I'm hoping, I did not screw up in too many places. There really are a _lot_ of examples in Mozzi!

Further note that I still have some plans to modernize the "STEREO_HACK", again in a backwards compatible way. Some hints to that are already in this PR, but the rest, I intend to provide separately, later.

I adjusted only one of the DAC examples. @tomcombriat : I'd like to pick your brain on how to do this best:

### A) Resolution when using EXTERNAL_AUDIO_OUTPUT
I rather think that these examples should probably use automatic scaling, too, even if it continues to be possible to simply pass an int value. For this to work, we'll have to assume some internal resolution. Right now, I have set things so that the resolution set when using EXTERNAL_AUDIO_OUTPUT is assumed to be 16 bits. (Note that this is also the max size of regular ints on AVR!). Now this has two drawbacks: First, it may create some unneeded shifting (i.e. a minor performance hit). Second, and more importantly, this does not make it easily possible to support more than 16 bits, cleanly. Of course this could be turned into yet another config setting, aka ```#define EXTERNAL_AUDIO_BITS 24``` but I'm not sure this is the way to go.
 
### B) audioOutput() function signature
Right now, I have declared the external audio output function as ```inline void audioOutput(const AudioOutput_t f);```. This effectively translates into ```inline void audioOutput(const int f);``` for mono, and ```inline void audioOutput(const StereoOutput_t f);``` for stereo, where StereoOutput_t is a struct of two ints (declared in AudioOutput.h).
This may or may not be ideal. Now, for external audioOutput(), you'll have to know, whether Mozzi is configured to use stereo output or not (i.e. you'll have to add #ifs for STEREO_HACK). Ideally, I'd like to avoid that whole mess by encapsulating both stereo _and_ mono output in a struct. That would not impose any computational overhead, and it would allow to provide getter functions that could simply provide a way to retrieve a second (duplicated) channel even from a mono sample. Anyway, I don't see a way to do so, without breaking all existing sketches, so this isn't viable. Yet, long story short, what do you think is the best function signature to use for audioOutput(), long term?

tfry-git:
Hang on, I actually have an idea on part B of the above. Will have to work out some details and do some testing, first, though.

tomcombriat:
Hi,
I did not have time to go very deep in this new implementation, will probably do this weekend! However a few thoughts:

  - this will simplify lot of things on the long run, especially for people like me working on different platforms with Mozzi!
  - **about part B**: this structure for handling the audio channel(s) is a great idea and will ease a lot the ports for stereo on different platforms. `int` seems a good idea on the first glance so it adapts naturally to 32 bits platforms which are more likely to be able to output more than 16 bits. However, if we want to be as general as possible, maybe declaring this structure as a template with default type `int` would allow an advanced user to use `long` instead if needed to output more than 16 bits on an AVR, for example using two 12bits DAC as it is done in one example. I am not sure how worth it would be, and how a mess it would be to have this structure casted by the user. I am not yet very familiar with templates so there might be a trick that I do not know.
  - **about part A**: I would personally go for  the `#define EXTERNAL_AUDIO_BITS N_bits` solution: if a user is already modifying `mozzi_config.h` in order to use an external audio function, then he/she could set the bit as the same time. It does not imply much more configuration than already enabling the `EXTERNAL_AUDIO` capability for the user, while greatly simplifying the life of users using more bits or giving a (very) small boost of performance to the ones which are using less bits… I mean, it is not like asking to change something in `MozziGuts.cpp`! 16 bits could be put default though as this seems a good balance for most usages. Just to mention it: I actually managed to get more than that (approximately 20) out of the noise, so that's definitely not a dream to get more than 16 bits quality with mozzi.

I do not know if that was of any help, some ideas might come as my brain is now unconsciously thinking about it and also when I'll have a bit of time. As I understand, this PR will break all sketches that do not use the `AudioOutput::` signature or did I miss something?

tfry-git:
> As I understand, this PR will break all sketches that do not use the AudioOutput:: signature or did I miss something?

The great thing is that it won't. Existing sketches (except for those using EXTERNAL_AUDIO_OUTPUT) continue to work. I'm in the process of tossing a few more ideas around, and am quite happy with how those are turning out. So hang on a few more days - and sorry about the noise. But yeah, it simply takes time to think these things through, so I'm glad you're with me, unconsciously ;-).

tfry-git:
Ok, so here's my new take. Again, first thing to understand is that this will _not_ break existing sketches except for those using EXTERNAL_AUDIO_OUTPUT. It's all source compatible. "AudioOutput_t" still is an int (in a mono config), or a void (when using STEREO_HACK), although in the much longer run, that may actually be worth changing, too.

Also, despite the fact that the "new" updateAudio() looks rather different, that's all syntactic sugar (and no fat!) that the compiler will burn through to arrive at the same assembly as before on AVR (verified for -Os, -O1, and -O2), only taking care of appropriate shifting, automagically.

Let's look at some examples of what it boils down to from a user perspective (*now* would be a good time to provide feedback on the naming of those functions and structs!). I'll hint at STEREO_HACK in a few places, although actually, I do plan on deprecating (not removing) that hack in favor of a cleaner solution in a separate PR after this one, but that will work without any further changes to updateAudio().

```
int updateAudio(){
  return oscil.next();
}
/* becomes */
AudioOutput_t updateAudio(){
  return MonoOutput::from8Bit(oscil.next());
}
```

Ok, I agree that is rather more verbose, but it is also more explicit on what is being done. Also, it will work in HIFI mode, and on e.g. STM32 without any adjustment.

To see the beauty of it, let's look at what becomes of the HIFI version of the above example:

```
int updateAudio(){
  return oscil.next() << 6;
}
/* becomes */
AudioOutput_t updateAudio(){
  return MonoOutput::from8Bit(oscil.next());
}
```

Yep. That's just the same thing. HIFI or not becomes a configuration detail, not something to design into your sketch.

On to another common example:

```
int updateAudio(){
  return ((int) oscilA.next() * oscilB.next()) >> 8;
}
/* becomes */
AudioOutput_t updateAudio(){
  return MonoOutput::from16Bit((int) oscil.next() * oscilB.next());
}
```

Here, too, figuring out the correct shifting to use is the job of the compiler, now.

On to where it gets a bit messy: Sketches pushing the limits of the STANDARD_PLUS output resolution:

```
int updateAudio(){
  int asign = (int) oscilA.next() + oscilB.next();
  return (constrain(asign, -244, 243));
}
/* becomes */
AudioOutput_t updateAudio(){
  int asig = (int) oscilA.next() + oscilB.next();
  return MonoOutput::fromAlmostNBit(9, asig).clip();
}
```

I'm rather proud of the ```.clip()```, replacing the constrain(), again in a way that will just work across all configs. I'll admit that I'm not so proud of the naming of ```fromAlmostNBit()```. Perhaps you have an idea on that? ```fromAlmostNBit(n)``` does the same thing as ```fromNBit(n)``` on all platforms _except_ STANDARD_PLUS, where the former will shift (if needed) to 9 bits range, while the latter will shift to 8 bits.

Now let's peak at stereo. First the state with the current STEREO_HACK, then the form that I have in mind for later (again, keeping backwards compatibility!):

```
void updateAudio(){
  audio_out_1 = oscilA.next();
  audio_out_2 = oscilB.next();
}
/* becomes (optionally) */
void updateAudio(){
  StereoOutput out = StereoOutput::from8Bit(osilA.next(), oscilB.next());
  audio_out_1 = out.l();
  audio_out_2 = out.r();
}
/* will become */
AudioOutput_t updateAudio(){
  return StereoOutput::from8Bit(osilA.next(), oscilB.next());
}
```

With the latter, it will be even trivial to make sketches written for mono compile on stereo (with a warning), and vice-versa.

### What about EXTERNAL_AUDIO_OUTPUT?

@tomcombriat Ok, so I'll go with EXTERNAL_AUDIO_BITS. But for the appearance of the audioOutput() function, that is now:

```
void audioOutput(const AudioOutput& f) {
  myDAC.write(f.l());
  myDAC.write(f.r());
}
```

Mono, stereo? No need to worry in the output function (same as now). Although of course it's also possible to write:

```
void audioOutput(const AudioOutput& f) {
  myDAC.write(f.l());
#if (STEREO_HACK == true)
  myDAC.write(f.r());
#endif
}
```

The difference to the current output signature is that in case of mono, hidden behind the "AudioOutput" label only a single int is passed, rather than two. But also, all of this will make it a trivial excercise to expand to 5.1 or whatever, if that wish should arise one day.

tomcombriat:
Hi @tfry-git !

That's great! I'm looking forward to test that! Most of my sketches will be broken as I use external_audio nearly always but I think that worth it! Maybe a post on the forum will be worth it as not so many people have a peak on PR discussion. I know that you do not look google forum (me neither actually) but if you do not want to create an account I can take care of that if that helps. That will be the occasion to present the new DAC features also.

An idea just occurred to me (unconsciously): as it is now, the user sometimes need to dig into mozzi_config.h, even for running certain examples (especially when using an external audio output function). Would it be worth to add a preproc variable `#define ALTERNATE_CONFIG` for instance that, if defined at the beginning of the sketch, won't load the config from there, but instead from a config file defined by the user in the same folder than the sketch (which then would define everything that is defined in mozzi_config.h and which would need to be included before mozzi in the user sketch): just making possible to *not* load mozzi_config.h, basically, so that it won't overwrite a user #define. This way, the sketches and the examples would run "out of the box" if an alternate configuration is needed and would not need anymore to ask the user to make changes in the "default" mozzi_config.h which is something which has puzzled me for some time: 24bits examples for instance would define the EXTERNAL_AUDIO_BITS in such a file, alongside the USE_EXTERNAL_AUDIO. Do you think that would worth it?

tfry-git:
> but if you do not want to create an account I can take care of that if that helps

That would be much appreciated! I'm trying to keep my balance, somewhere, and it would be cool, if I would not have to give up that arbitrary line marking how deeply I wanted to become involved.

----

> as it is now, the user sometimes need to dig into mozzi_config.h

Yes, that has been bugging me for quite some time, too. Using a preprocessor variable probably won't help, as the library itself is compiled in several separate compilation units (e.g. MozziGuts.cpp is a separate unit, and includes all those headers, separately). Those would all need the same config, obviously. So all my own ideas along those lines ended in nothing.

*However*, I think [__has_include()](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html) may be something that could save us. With that, it should be possible to include a config from the sketch folder, if it exists. I have not started looking at details of how this could be done, though. The platform specific defines should ideally be integrated into that mechanism, somehow, too (and possibly re-organized).

tfry-git:
At the danger of adding too much noise to this discussion, I'll toss in an idea:
```MonoOutput::fromNBit(bits, data);``` and ```MonoOutput::fromAlmostNBit(bits, data);``` could also be named as ```MonoOutput::scale<bits>(data);``` and ```MonoOutput::squeeze<bits>(data);```, respectively.

Personally, I think the template syntax ("<bits>") would be nice to read, but maybe they look scary to some? The shorter labels are just a thought, and I have no personal preference on those.

tomcombriat:
> and it would be cool, if I would not have to give up that arbitrary line marking how deeply I wanted to become involved.

No problem, I unfortunately gave that up some time ago (for Mozzi I think :/), That's unfortunate but now that I have it I can "carry the burden" for some others ;).

> Yes, that has been bugging me for quite some time, too. Using a preprocessor variable probably won't help, as the library itself is compiled in several separate compilation units (e.g. MozziGuts.cpp is a separate unit, and includes all those headers, separately). Those would all need the same config, obviously. So all my own ideas along those lines ended in nothing.

Ah, yes, those damn compilation units, I remember you already answered that to one of my suggestion… I did not know this function, I am always afraid that this ends up a bit hacky but that might worth it, in the long term…

> Personally, I think the template syntax (<bits>) would be nice to read, but maybe it will look scary to some? The shorter labels are just a thought, and I have no personal preference on those.

Aren't templates supposed to be faster? I now see that you used macros in there so I guess both solutions are binary equivalent (I am fairly sure that you know these kind of things way better than me, so that's more a curiosity question…). If that can help the choice, there are already templates in Mozzi, for envelopes, oscillators and so on, so I do not think this syntax is more scary for beginners. I have to say that I am not a big fan of `squeeze` and `scale` that I find to subjective compared to `fromNbit`. But that's no more than a personal opinion…

tfry-git:
> Aren't templates supposed to be faster? I now see that you used macros in there so I guess both solutions are binary equivalent [...]

Exactly. Templates would be faster, if anything would actually be left of the function call, but the compiler can inline all of it (and I checked that it really does). Good point about templates being already used across Mozzi.

tomcombriat:
Hi again!
I was wondering how much this is in working state already? I was thinking of cloning it, test it and probably adapt the remaining examples in the next few days but there is little point if you are still "deep in". I'm looking forward to test that!

tfry-git:
Hi!

This is in a working state. Or at least I hope so, as I only tested so much, and then changed more stuff after that. At any rate, testing is essentially what's left to be done.

Changes that might still happen are the naming (but not the behavior) of those functions, as discussed. Also, I'm not yet 100% sure, whether it's better to pass the AudioOutput parameter as a reference (the "&") or as a plain value in audioOutput().

tomcombriat:
Hi!

Just starting testing, with a particular focus on the `EXTERNAL_AUDIO_OUTPUT` capability.
I have a few compilations errors, even with the example you modified: trying to compile it I get: `request for member 'l' in 'f', which is of non-class type 'const int'`. So apparently for mono sketches Mozzi considers `f` as a `const int`, without any member. Changing the output line to:
`  uint16_t out = (f >> 4) + 2048;` instead of `uint16_t out = (f.l() >> 4) + 2048;` seems to resolve this problem but I guess that was not your goal as this changes the looking of this function…
However, some extra errors appears, which seems to relate to the "standard output" using PWM:

`
/home/tom/Arduino_install/arduino-1.8.4/libraries/Mozzi/MozziGuts.cpp: In function 'void startAudioStandard()':
/home/tom/Arduino_install/arduino-1.8.4/libraries/Mozzi/MozziGuts.cpp:553:11: error: 'AUDIO_CHANNEL_1_PIN' was not declared in this scope
   pinMode(AUDIO_CHANNEL_1_PIN, OUTPUT); // set pin to output for audio
           ^
/home/tom/Arduino_install/arduino-1.8.4/libraries/Mozzi/MozziGuts.cpp:560:38: error: 'PWM_RATE' was not declared in this scope
   Timer1.initializeCPUCycles(F_CPU / PWM_RATE,
                                      ^
/home/tom/Arduino_install/arduino-1.8.4/libraries/Mozzi/MozziGuts.cpp:564:14: error: 'AUDIO_BIAS' was not declared in this scope
              AUDIO_BIAS); // pwm pin, 50% of Mozzi's duty cycle, ie. 0 signal
              ^
`

Mozzi is seeking for Hardware definitions that are not available anymore as, using an external audio output function, no `AUDIO_CHANNEL_1_PIN` should be defined… Any thoughts about that? I tried to find the errors but it always takes me a few days to get my head back into MozziGuts… I however think that:

`#if (STEREO_HACK == true)`
`void updateAudio();`
`#else`
`AudioOutput_t updateAudio();`
`#endif`
should become
`AudioOutput_t updateAudio();` only, as the structure is defined for both stereo and mono. It compiles both ways anyway so that's probably minor…

tfry-git:
Thanks! Indeed I had not tried that combination at all. Before my PR, the default audio config header (i.e. usually AudioConfigStandardPlus.h) would still be included, along with the timer setup. I had changed that, but obviously I had only half-way through. (I remember to have made appropriate adjustments for several arches; but now I wonder which other ones I forgot).

Anyway, AVR now compiles with EXTERNAL_AUDIO_OUTPUT for me. Different to what we had before, this no longer setup up any pins, and only sets up Timer 1 at AUDIO_RATE (before, as long as STANDARD_PLUS was set, it was running at twice the audio rate, by default). As you said, I have done this using a preprocessor test in MozziGuts.h.

Also, AUDIO_BIAS is now defined according to the EXTERNAL_AUDIO_BITS setting. (MozziGuts.h), so AUDIO_BIAS should now be always available, if needed.

> I get: ```request for member 'l' in 'f', which is of non-class type 'const int'```

My usual sloppiness. I confused AudioOutput_t and AudioOutput in the example (which I hope may eventually become the same thing - in the much longer term - but currently cannot due to STEREO_HACK). Fix comited.

Also, I applied your suggested simplification in MozziGuts.h.

Now, I'm waiting anxiously, what other embarrassing mistakes you are going to find...

tomcombriat:
Hi, thanks for this!

I have found something else haha! For STM32, if `EXTERNAL_AUDIO_OUTPUT` is true, then `AudioConfigSTM32.h` is not included in MozziGuts. This file mainly defines output pins and so on, which are not needed for `EXTERNAL_AUDIO_OUTPUT`, but it also defines a few timers: `AUDIO_UPDATE_TIMER` which is needed nevertheless, and `AUDIO_PWM_TIMER` which is needed at line 359 of MozziGuts.cpp. The last one might need to be surrounded by `#if (EXTERNAL_AUDIO_OUTPUT != true)` // `#endif` if I'm correct?

I do not know if you prefer me to commit pull requests on this PR (that's a bit meta…) or just comment here… I can test unfortunately test only on Teensy, STM and AVR, and focusing only on STM. I am a bit afraid that I'lI miss "general" changes that have to be made on all platforms as I do not have a general view on this yet…

tomcombriat:
Hi again!

Okay, I got the example with the PT8211 DAC stereo working, I was a bit mislead by the example of the stereo_hack:

  - `int audio_out_1, audio_out_2;` should not be defined in the sketch right?
  - `return StereoOutput::fromNBit(24, (long)pan*asig, ((long)65535-pan)*asig);` is commented in the example you modified for the STEREO_HACK, so we should use the old way? By `assigning audio_out_1/2` in `updateAudio()`? It is working like that here, but I also had to change the return type of `updateAudio()` from `void` to `AudioOutput_t` even though no return value is needed?

tfry-git:
> I think I have found something else haha!

Yup, but at least this time, I manged to push the fix, before you finished typing ;-). I seem to have made the inclusion of all those config files conditional some time late, and apparently, I did not even test compiling with EXTERNAL_AUDIO_OUTPUT afterwards. Shame on me! Anyway, I believe I now got all those fixed.

> I do not know if you prefer me to commit pull requests on this PR (that's a bit meta…) or just comment here

I don't really know a good workflow for collaboration on PRs. Either would be fine with me, but I guess while it's small-scale fixes, commenting will be easier.

> I can test unfortunately test only on Teensy, STM and AVR, and focusing only on STM

Yes, the whole trouble is that the testing matrix has grown huge. I'll do my best to get the ESPs working decently. For SAMD21, I can at least test the compilation, even if I don't have the hardware. At least we're moving towards _a little_ better modularity, here.

tomcombriat:
> I don't really know a good workflow for collaboration on PRs. Either would be fine with me, but I guess while it's small-scale fixes, commenting will be easier.

Okay, I'll continue like that for the main core (I'm still not very confident with modifying Mozzi_guts… The modularity is for sure way better but I do not have your overlook), but will PR for the examples if that helps!

> Yup, but at least this time, I manged to push the fix, before you finished typing ;-).

I think I typed some extra in the meantime, while you were typing this sentence ;), it contains questions about the STEREO_HACK…

Sorry, I should probably have been more involved in the implementation, that's why I'm not very confident in modifying things now…

tfry-git:
About STEREO_HACK: My bad, once more. Fix committed. TLDR: Leave stereo as is, for now.

Long version: STEREO_HACK is - well - a hack. Unfortunately, that hack is not really fixable, without breaking existing sketches. However, the basic plan for that will be to create a new config define for stereo. When that is set, updateAudio() will indeed return AudioOutput_t for stereo as for mono, and audio_out_1/2 will no longer be needed. When STEREO_HACK is set, instead, a warning message will be shown, but the sketch will continue to work as before.

You _could_ change the void for AudioOutput_t, already, but there's no real gain in that, right now (and it wasn't supposed to be required, in the first place).

tfry-git:
As a side-note: The platform-specifics left in MozziGuts.cpp are mostly about

a) setting up audio output (pin modes, pwm timers, etc.)
b) setting up a timer running at audio rate (for configs that need it)
c) timer ISRs for the platforms that need them

It should really be possible to sort that stuff in a much more readable way. But that, too, is for another day...

sensorium:
Hi Thomas, Tom,

regarding testing, when I was testing regularly I made a script to compile and record all the examples.  I could come back later and listen quickly to find any problems.  Don't know if it's any help..

sensorium:
Here's the makefile if it's any use to you...

tomcombriat:
Hi!

> You could change the void for AudioOutput_t, already, but there's no real gain in that, right now (and it wasn't supposed to be required, in the first place).

I think that is a good idea to do like you did in the example to "prepare" the transition. Thanks! I'll soon commit a few changes on the examples!

@sensorium : thank for the comment! I personally did not received the attachment unfortunately :/

tfry-git:
@sensorium That makefile sounds really useful, indeed. But github appear to have scrubbed it (I did not get anything, either).

@tomcombriat I merged your branch into this one. By now I'm 95% certain that using the signature ```void audioOutput(const AudioOutput f)``` will be a tiny bit better than ```void audioOutput(const AudioOutput& f)```. Not much of a difference, but while I'm at breaking this, I guess I should try to get it right. Will look at the assembly a bit more, with a fresh head. Anyway, the only thing that will change is removing the "&". I'll take care of that, if it's needed.

tomcombriat:
Hi!
@sensorium Just a suggestion, maybe this Makefile could be placed in the "extras" folder? It might not be useful to regular users but developers might be happy to found it!

@tfry-git Out of curiosity, I would be curious to know why removing the `&` is a bit better. I remember from my long time ago (and quite low level) C++ courses that passing using references would only pass a single pointer whereas passing the whole structure (in the stereo case) would make a copy and a passing of two elements. Probably the fact that this structure is declared as `const` actually enables a possibility for inlining the whole function. I am just curious :). Anyway, I am sure a well chosen `sed` command should do the job ;).

sensorium:
I've added extras/scripts.  The makefile I used was https://github.com/sudar/Arduino-Makefile, but there are now lots to choose from for different platforms if you search Github.

tfry-git:
> Out of curiosity, I would be curious to know why removing the & is a bit better.

In our case, there is not a whole lot of data passed around (not even in the stereo case), and the reference does induce some overhead(*). It is equivalent to passing a pointer: First an address is taken, this is passed (usually taking exactly just many bytes as an int), then in the function the value is read from that address (which, in the case of stereo will also be several reads).

Using references is clearly superior, when:
1) Not all of the data passed is actually used (passing a struct that contains information that isn't relevant for the function called).
2) The data should be modified in place.

Further, using references induces no overhead, when:
3) The data to be passed is a variable (lvalue) OR it is stored in a memory location that the compiler is sure will not be changed _indirectly_ during the call to the function. Then passing a pointer to that is all the compiler has to do. If the compiler cannot be sure of that, it will instead create a copy of the data to be passed, after all (or at least, that is my understanding, now).

I wasn't so sure about 3 (and I'm still not sure, my interpretation is correct, but the observed effect is clear, anyway). The relevant line in MozziGuts is: ```audioOutput(output_buffer.read());```. Would this simply be able to pass the address of the current sample inside the buffer? I believe, then, the overhead would simply fall away. But no, apparently not so. The compiler seems to go to quite some lengths, first copying the value from the buffer, then pushing the address of that copy to the stack (which is how function parameters are passed). Then, in audioOutput(), the address is popped from the stack, and then the value is read from that address. I guess the reason for that copy may be that the compiler - not being able to see what's inside audioOutput() - cannot rule out that the output_buffer is overwritten by something done inside audioOutput(), thus making a copy, first.

In contrast, in pass-by-value, the current sample taken from the buffer is simply pushed to the stack, then popped inside audioOutput().

(And if/where audioOutput() could be inlined, the pushing and popping could be omitted, altogether, regardless of the function signature.)

BTW, should you want to experiment with this, you can use "__attribute__ ((noinline))" to keep the compiler from inlining, even inside a single sketch. The sketch size generated is a first rough measure. To peek at the assembly, take a note of the temporary dir shown in the compiler messages, then call ```avr-objdump -S MySketch.ino.elf``` to view the result. The "-S" will mix the source lines between the assembly, so you'll have some orientation. In my tests, passing by reference caused 32 bytes of additional instructions, which may not even be all that insignificant on AVR.

So I'll update in a while, tomorrow, or the weekend.

------

(*) If it isn't inlined, in which case there is no difference at all. However, that will not (usually) be the case for EXTERNAL_AUDIO_OUTPUT, as audioOutput() will again be in a separate compilation unit, then.

tomcombriat:
Ok, thanks! That's very interesting! I would have bet the opposite from the courses I had but I guess C++ for scientists is not most complete course for these kind of low-level optimizations… When I do not have valgrind I am a bit lost…

Anyway, will soon try to test more, on Teensy (with and without EXTERNAL_AUDIO_OUTPUT) and I would say this long story is getting very close from being done!

tfry-git:
Alright, I pushed, what I hope to be the final iteration, then. (Replacing AudioOutput& with AudioOutput).

If there are on reported catastrophes, I'll merge into master on Monday.

tomcombriat:
Just in case you did not see it, I pulled a few minor corrections for this PR in #111 .

If a small announcement is made in the forum, should I or @sensorium take care of it?

All the best in these times…

sensorium:
Hi Tom, Thomas, 
feel free to announce your work!

It's a significant addition to Mozzi.

Do you have any thoughts about licensing?

Tim

tfry-git:
Hi Tim,

regarding licensing, I wrote some years ago that, essentially you can use my contributions as you please. That still stands. Personally, I'm sympathetic to the idea of dual-licensing under GPL 2+, as requested, here: https://github.com/sensorium/Mozzi/issues/72 . That would give open source users a few more options (they could chose whether to use CC-BY-NC-SA or GPL in their projects, allowing to combine with more libraries such as ESP8266Audio). But again, your call, entirely.

Thomas

tomcombriat:
Youhou! That's the end of a long story! Thanks @tfry-git :)!

Just to mention that I would also be sympathetic for a dual-licensing but, unlike Thomas I am not one of the main contributors to this project so you should not take my worlds with too much weight.

Best,

PR #99, opened 2020-11-29
Use higher audio rate by default on faster boards

tfry-git:
AUDIO_RATE 16384 is a sensible default for the classic Arduino boards, but for the newly added platforms (Teensy3, Arduino Due, STM32, ESP8266, soon ESP32), that does not seem to make a lot of sense.

This PR simply ups that to 32768 on these platforms, while leaving AVR unchanged. With this, there should be a little less need to customize mozzi_config.h with. The PR might affect sketches that are already pushing the stronger boards to the limits, but I think that risk is mostly theoretical.

One point for discussion could might be, whether it would make sense to go even higher, though. Thoughts?

### DIFF
--- a/hardware_defines.h
+++ b/hardware_defines.h
@@ -29,4 +29,10 @@
 #define NUM_ANALOG_INPUTS 1
 #endif
 
+#if IS_AVR()
+#define AUDIO_RATE_PLATFORM_DEFAULT 16384
+#else
+#define AUDIO_RATE_PLATFORM_DEFAULT 32768
+#endif
+
 #endif /* HARDWARE_DEFINES_H_ */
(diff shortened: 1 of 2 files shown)


tomcombriat:
Hi!

I think this is a great idea! I personally keep my config on 32768 but only using STM32 boards. I always wanted to test higher, but as I was satisfied with 32k, and already pushing the boards to the limits, I did not find the motivation.

I would say that the potential affects on sketches that are close to the limit are of very little concern compared to the benefices of this PR.

dietervandoren:
I've been working a bit on ESP32 in the last few days. The auto rate selection in this pull seems to work fine. 

> One point for discussion could might be, whether it would make sense to go even higher, though. Thoughts?

I'm sorry, not experienced enough with Mozzi and different processors types to make a judgement.

PR #103, opened 2020-12-28
External audio: added PT8211 16 bits dual DAC support

tomcombriat:
Hi!

This adds two examples expanding Mozzi's capability to use external DACs using an user defined `audioOutput` function.

This chip is a stereo (2x16bits) DAC that is normally interfaced by I2S, which is emulated here using SPI and should work on any platform supporting this protocol (tested on STM32 and AVR).
There are two examples:

  - PT8211_stereo_16bits is a minimal example working for all platforms with SPI using the default SPI port.
  - PT8211_stereo_16bits_STM32_SPI2 is a specific example using the SPI2 port on the STM32 thus freeing the ADC pin located on the default SPI port. This might not be relevant here but originates from a long discussion with another Mozzi user trying to figure out how to make this port work.

This might conflict with #98 but is coherent with the previous implementation on the stereo Hack. After testing #98 I'll probably move all the stereo examples to the new paradigm *as a whole at the same time*, maybe so that it will be pull request at the same time as the new implementation.

Any comments are appreciated!

Best,
Tom


tfry-git:
I didn't know, this was possible. Nice! I haven't done any testing, but that looks like a wonderful option, in particular for STM32.

You are right, both examples will need a small adjustment after PR #98, but that will be trivial, and like you say can be done independently.

So, no objections. Some small things I noted: Is it possible to use SPI.transfer16() instead of two SPI.transfer()s? Or does the whole thing rest on the behavior of the CS-pin in between the transfers, or some such?

The PT8211 takes data in LSB format, which, I believe, is a bit uncommon. Now, we cannot show everything in a single example, but perhaps it would make sense to state as much in comments, and also what parts would need to be changed for MSB bit order.

tomcombriat:
Hi! Thanks for your comments!

Yeah, I was a bit sad when I figured out the I2S was not on the BluePill as everybody was speaking about the PT8211 chip… But it works by emulating it with SPI! I2S can be used asynchronous or synchronous, which helps a lot. And using the second SPI on the BP really change the things and makes it a really great platform for Mozzi! 

Actually, the PT8211 is Least Significant Bit Justified, but quoting the datasheet: "The DIN data must be the 2’s complementary format and the MSB (Most Significant Bit) must be the first." This is why the SPI configuration takes MSBFIRST as argument. The biggest difference from what is usually done in Mozzi is that it works with the 2's complementary (zero centered) but this is not a problem as this is how Mozzi is working internally.

I changed the calls for `transfer()` to `transfer16()` as this is indeed more logical (I always forget about this function… At least I did not bitbang it!).

Attached a quick and dirty scope of what it gives!

Best and happy new year!
IMG20201231111857

PR #111, opened 2021-01-24
Added missing bracket

tomcombriat:
This fixes one example. Will do a few more test (if I can put my hand on this Teensy chip hiding somewhere…)


tomcombriat:
On top of #98

tomcombriat:
Added the value of AUDIO_BITS which was missing for Teensy.

tomcombriat:
So, in the end, #98 seems to work with Teensy (tested one example with native DAC, and one with EXTERNAL_AUDIO_OUTPUT).

PR #113, opened 2021-01-27
Deprecate STEREO_HACK

tfry-git:
Ok, so this is what I had in mind for modernizing stereo output. STEREO_HACK continues to work, but generates a compiler warning.

My main concern is really, whether the defines are named well. Are there foreseeable clashes with other libraries, for instance? Should we just use "1" and "2" for MONO and STEREO?

As another point for discussion: Sketches compiled in the "wrong" mode (mono in a stereo config and vice versa) will now produce a compiler warning, but will work, otherwise (duplicating the mono channel, or using only the left channel of stereo). Would it be preferable to break compilation, as a stronger hint to users?

Status: I have not done anything resembling real testing, yet. All I tried was compiling one example, and only on AVR, so far. Other than that, the above concerns, and whatever other comments you may have, I consider this PR "ready". The remaining ocurences of "STEREO_HACK" in the code are places meant to provide backwards compatibility.

Note: For the gist of it, look at the updated examples.


tomcombriat:
Hi!
 
If that helps, just some quick comments:
    - as it is now, I do not see any clashes for the `defines`, at least from the libraries I am using.
    - I would be to break the compilation if the sketch is compiled on the wrong mode: depending on the included libraries, a lot of warnings can be prompted during compilation, and users might overlook them as it would still work. Breaking the compilation would "force" a bit the users to both have a look in `mozzi_config.h` (which is always nice) and to understand a bit what is going on there…

Nit: in the news I spotted a mispelling: 
> updateAudio() should be slightly rewritten for portability across different **configuraitions** and boards.

Also, just to mention it, the STEREO_HACK is currently working on STM32 when using an EXTERNAL_AUDIO_OUTPUT" but I do not know if that is worth mentioning…

Best,

pschatzmann:
What was the reason why you deactivated the automatic conversion ? 
I would have really liked that feature - because all existing sketches would automatically work w/o giving any compile errors and give output on 2 channels.

ps. what is the timeframe for a merge to master ?

tfry-git:
> What was the reason why you deactivated the automatic conversion ?

That was the feedback I had from @tomcombriat . The idea being that this makes it easier to spot accidental misconfiguration. I'm open to both, really.

> what is the timeframe for a merge to master ?

If you have no objection WRT to merging what we have right now, this is ready to go in (and overdue), and I could do the merge tonight. I still promised to do some generalizations along the lines of your MultiChannelOutput, so that might potentially be the holdup. OTOH, I believe it should be possible to add that, separately, and the re-implement MonoOutput and StereoOutput on top of that, afterwards.

pschatzmann:
I would suggest to move ahead with what we have, and take it from there.
Next week I will try into integrate this into my fork...

PR #120, opened 2021-04-09
Changed comments suggesting the pitch being encoded as frequency

tomcombriat:
Hi,
Reading the doc, I was surprised that it was written there that the PDresonant filter was receiving the frequency of the note as parameter, encoded in `byte`. If I am correct, it actually expect the midi note value.

Not a big change, but can maybe remove some confusion.

Best,
Tom


sensorium:
Thanks Tom

PR #128, opened 2021-06-12
How manually override frequency selection for STM32 inside Arduino

pirvcb:
I beg help for this.

https://groups.google.com/g/mozzi-users/c/svJcWE2mws8

Thanks and pardon for unusual approach...


pirvcb:
wrong place

Issue #132, opened 2021-08-07
16 bit wave table support


daanklijn:
Currently the wave tables have an 8 bit resolution. It would be nice to also support a higher resolution like 16 bit, especially in combination with a 12 bit DAW. Any idea how to implement this?

daanklijn:
I managed to quickly add 16 bit wavetable support by slightly altering the `Oscil` class. My rough implementation can be found in the `Oscil16Bit` class. Although it might be better to include this functionality directly into the `Oscil` class / let the `Oscil16Bit` inherit from `Oscil`, I currently don't have the time / Mozzi knowledge to properly implement this.

donluca:
I would definitely love to see this extended to all the other wave forms.
Now that powerful devices like ESP8266 and ESP32 are around, using 16-bit wave forms should be a non-issue in most cases.

eclab:
I would also like to see this, particularly for the AVR arduino.  Yes, it's 8-bit, but its output range is 0...488 and we can't take advantage of that right now with Oscil: in effect, from16Bit isn't all that useful because the input was just upsampled from 8 bit!  Perhaps we could make an Oscil16?

sensorium:
I hope this doesn't sound defeatist, but as the initial author of Mozzi, I think at this time that to make meaningful use of 16 bit wavetables, most of the library would need redesigning.  
The main challenge in first writing Mozzi was optimising it within the limitations of the 8 bit processors which were available then.  Even if you have a 16 bit wavetable, you'd often need to scale it back to 8 bits to work with the synthesis classes in Mozzi.  
I might be wrong, but I think the code would become unwieldy if you tried porting all of the synthesis classes differently for multiple processors.  If you want higher resolution synthesis, I expect it would be simplest to find or begin a new library designed for more capable processors.

tomcombriat:
Hi,
Just to add a bit of fuel to this discussion: I actually think that most of the classes would manage more or less okay, especially the ones that have been modified recently where we tried to take 32 bitters (the `ResonantFilter.h` is a good example for instance). Now, as @sensorium said, some others might need the same type of modifications but that should be more or less limited to the ones dealing directly with waveforms (can't think of any on top of my head). There might also be a bit of the backbones to modify but recently with the modularization of `AudioOutput` should be quite straightforward (I managed to output more than 16 bits on some designs).

[...]

PR #133, opened 2021-08-07
16 bit wave table support

daanklijn:
See #132


tomcombriat:
Hi!
Sorry it seems this went a bit under the radar…

@sensorium and @tfry-git any opinion about this? Considering the re-use of code, maybe this can be done in the main `Oscil` class.
I have to say that I never really saw the need of 16bits oscillators as, multiplying an oscillator to another one, or an envelope, already leads to a resolution usually greater than the output capability of many systems. But I might miss some cases!

sensorium:
Maybe if the library evolves enough ;) , 16 bits could be useful... 
@daanklijn, if you're around, have you used this in a sketch?  Since many processes in Mozzi still expect 8 bit inputs, I don't have an immediate idea of what you could do with 16 bit sources. So an example would be great.
If the code works and could be added to Oscil without complication, I'd be open to it...

poetaster:
I don't know if this is directly related but, I've been looking at porting the 22050 hz wavetable header files (22khz 16 bit signed PCM format ) from https://github.com/poetaster/scarp/tree/main/firmware/Pikobeats into a mozzi context (this is a hardware/software project with multiple types of firmware). I'm not sure the bit depth is an issue at 22050 hz, but if I got it correctly, the current wavetables are limited to 8bits and 16khz?

I'll try converting the some of the sample files to the mozzi format and see how it sounds (PWM on an pi pico).

tomcombriat:
> but if I got it correctly, the current wavetables are limited to 8bits and 16khz?

Well, the wavetables do not have a limiting frequency per se but indeed the sampling rate of the output is 32k maximum, hence giving a bandwidth up to 16kHz. There have been a few tries to double that on performant platform but was not finished).

And yes, so far the wavetables data are 8bits.

PR #137, opened 2021-10-03
Fix compilation for Teensy3

tomcombriat:
Mozzi would not compile with the new TeensyDuino (probably since the arrival of Teensy4). This is caused by a small change in the Teensy API regarding ADC.

This PR fix the problem and allow to use Mozzi on Teensy 3X with the newer Teensyduino.
Note that this will probably won't compile for people still using the old Teensyduino. In that case a simple update of Teensyduino solves the problem.

### DIFF
--- a/MozziGuts.cpp
+++ b/MozziGuts.cpp
@@ -396,8 +396,8 @@ void samd21AudioOutput() {
 static void startAudioStandard() {
 
 #if IS_TEENSY3()
-  adc->setAveraging(0);
-  adc->setConversionSpeed(
+  adc->adc0->setAveraging(0);
+  adc->adc0->setConversionSpeed(
       ADC_CONVERSION_SPEED::MED_SPEED); // could be HIGH_SPEED, noisier
 
   analogWriteResolution(12);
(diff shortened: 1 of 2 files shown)


tomcombriat:
This also fix a frequency shift because of a rounding error in timer calculation.

tfry-git:
Is it possible to check whether people have the old or new version of TeensyDuino?

tomcombriat:
@tfry-git Not of my knowledge yet. Will let you know if I stumble into it. If that in the case, I would suggest putting a clear error message at compilation asking to update Teensyduino to avoid spilling further `MozziGuts` into different sub-teensy cases. Also, I suspect audio_input to be impacted, so let's just wait some time before considering merging.

The rounding error was present for both versions of Teensyduino. An easy fix as the intervaltimer supports floating numbers.

I am now trying to port Teensy4X (hence the update of Teensyduino). So far everything is very close to Teensy3, so I am using the same audio config file to avoid too much code duplication. If that works, might be worth switching this file to AudioConfigTeensy. I am standard PWM as default output mode, except some dropouts as regular intervals, which I suspect being a wrongly configured PWM frequency it seems to work.

tomcombriat:
Will be superseded by #138

PR #138, opened 2021-10-04
Teensy4x port

tomcombriat:
This adds supports for Teensy4.

As the Teensy4 do not have any DAC, the default output is PWM on pin A8 (chosen for the lack anything useful on this pin for audio applications, who needs CAN anyway ;) ). Obviously, because of the capacity of this chip (800MHz base frequency, more than 1GHz on overclock!) it is a bit sad to use it with PWM but that gives a default "bare chip" output capacity.

Output has been tested, a few things remain to be tested before even considering merging:

  - external audio output
  - audio input (probably to be solved along with #137 if there is a problem there )
  - stereo (if that was in the original Teensy port).

I am personally looking forward for the sounds that can created with that much CPU power!

Best,

### DIFF
--- a/AudioConfigTeensy3_12bit.h
+++ b/AudioConfigTeensy3_12bit.h
@@ -1,6 +1,7 @@
 #ifndef AUDIOCONFIGTEENSY3_12BIT_H
 #define AUDIOCONFIGTEENSY3_12BIT_H
 
+#warning If you get a compilation error you should probably update Teensyduino to its latest version
 
 /** @ingroup core
 */
--- a/AudioConfigTeensy4_10bitPwm.h
+++ b/AudioConfigTeensy4_10bitPwm.h
@@ -1,0 +1,17 @@
+#ifndef AUDIOCONFIGTEENSY4_10BITPWM_H
+#define AUDIOCONFIGTEENSY4_10BITPWM_H
+
+#warning If you get a compilation error you should probably update Teensyduino to its latest version
+
+/** @ingroup core
+*/
+/* Used internally to put the 0-biased generated audio into the centre of the output range (10 bits on Teensy 4 using PWM) */
+#define AUDIO_BIAS ((uint16_t) 512)
+#define AUDIO_BITS 10
+
+#define AUDIO_CHANNEL_1_PIN A8
+#define AUDIO_CHANNEL_2_PIN A9
+
+
+#endif        //  #ifndef AUDIOCONFIGTEENSY4_10BITPWM_H
+
--- a/MozziGuts.cpp
+++ b/MozziGuts.cpp
@@ -27,7 +27,7 @@
 #if IS_AVR()
 #include "FrequencyTimer2.h"
 #include "TimerOne.h"
-#elif IS_TEENSY3()
+#elif (IS_TEENSY3() || IS_TEENSY4())
 // required from http://github.com/pedvide/ADC for Teensy 3.*
 #include "IntervalTimer.h"
 #include <ADC.h>
@@ -49,7 +49,7 @@ uint16_t output_buffer_size = 0;
     "Mozzi has been tested with a cpu clock speed of 16MHz on Arduino and 48MHz on Teensy 3!  Results may vary with other speeds."
 #endif
 
-#if IS_TEENSY3()
+#if (IS_TEENSY3() || IS_TEENSY4())
 ADC *adc; // adc object
 uint8_t teensy_pin;
 #elif IS_STM32()
@@ -150,8 +150,8 @@ uint8_t adc_count = 0;
 int getAudioInput() { return audio_input; }
 
 static void startFirstAudioADC() {
-#if IS_TEENSY3()
-  adc->startSingleRead(
+#if (IS_TEENSY3() || IS_TEENSY4())
+  adc->adc0->startSingleRead(
       AUDIO_INPUT_PIN); // ADC lib converts pin/channel in startSingleRead
 #elif IS_STM32()
   uint8_t dummy = AUDIO_INPUT_PIN;
@@ -170,8 +170,8 @@ static void receiveFirstAudioADC()
 */
 
 static void startSecondAudioADC() {
-#if IS_TEENSY3()
-  adc->startSingleRead(AUDIO_INPUT_PIN);
+#if (IS_TEENSY3() || IS_TEENSY4())
+  adc->adc0->startSingleRead(AUDIO_INPUT_PIN);
 #elif IS_STM32()
   uint8_t dummy = AUDIO_INPUT_PIN;
   adc.setPins(&dummy, 1);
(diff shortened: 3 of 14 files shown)


tomcombriat:
Thanks @sensorium , I like Mozzi and I am happy to try to contribute to it!

Since the last update of the documentation a lot of things have been modified but principally under the hood (#98 and the following). However, I think that the main page should at least reflect that it is now possible to use additional hardware to output the sound thanks to `EXTERNAL_AUDIO_OUTPUT` (#87 and #97 ) allowing for audio quality output if needed. There are a few open PR also that maybe should or should not be included before regenerating the docs.

> it seems ironic to use Mozzi on more powerful platforms since the main challenge initially was optimising it for 8 bit processing.

Indeed most of the PR these days are actually about keeping that optimization for AVR while not making it a limitation for more powerful platforms, sometimes through less optimized solutions. Actually that is what I really like about Mozzi, it can be used on a lot bunch of different platforms and output very different qualities of sound depending on what is the intended use, it is very flexible which is great! Obviously the bad side effect become clear when opening the guts of Mozzi but that is becoming easier and easier thanks to the modulation work of @tfry-git . Maybe this discussion should be in a new thread as you said. That would be interesting to have a place to discuss future developments and ideas for Mozzi!

AudioConfigTeensy4_10bitPwm.h
```diff
@@ -0,0 +1,17 @@
+#ifndef AUDIOCONFIGTEENSY4_10BITPWM_H
```
tfry-git, line 1: I have no idea what other output options would be "natural" on teensy 4, and how similar these would be to set up. But just as a point to consider. Could this be too specialized? Would it be better to use plain AudioConfigTeensy4.h, instead?

AudioConfigTeensy3_12bit.h
```diff
@@ -15,6 +16,8 @@
 #define AUDIO_CHANNEL_1_PIN A14
 #elif defined(__MK64FX512__) || defined(__MK66FX1M0__)
 #define AUDIO_CHANNEL_1_PIN A21
+//#elif defined(__IMXRT1062__)
```
tfry-git, line 19: What's this?

mozzi_config.h
```diff
@@ -97,7 +97,7 @@ compile time.
       MONO is 1 and the value of STEREO is 2, so future extensions are also expected
       to set this to the number of available channels. */
 #define AUDIO_CHANNELS MONO
-//#define AUDIO_CHANNELS STEREO
+ //#define AUDIO_CHANNELS STEREO
```
tfry-git, line 100: Nitpick, but please clean up that space, so it does not end up as a changed line.

tfry-git (commented):
Does your PR fully address the issue reported the other day (compilation with the new teensyduino, but also some trouble with analog reads)?

One thing left to do is be to add some documentation (README.md and extras/NEWS.txt).

Beyond this, I have not done any testing, but looks good to me.

tomcombriat:
Okaaaaaaay, I got a Teensy LC for testing! At first everything seemed to work alright, even `mozziAnalogRead` …

… until I tried to read more than one pin. Then, indeed, problems arise. This might also be the cases with the others actually, I have to say that testing several analogRead was not part of my testing procedure. Even before this commit I never used `mozziAnalogRead` on a teensy…

Will dig more and - hopefully - come with a fix soon.

tomcombriat:
I put my hand on something fuzzy:

For the Teensy (LC at least), `NUM_ANALOG_INPUTS` is equal to 13.
However, if I try to read the pin A4, the affected channel (`current_channel`) is equal to 18. This leads to an overflow of the `analog_readings` stacks (which is of size 13), because the pin number is directly used on the Teensy (`current_channel` = `pin`).

This overflow does not lead to any problem *in most cases - at least here - * when only one value is read: the value is written in the wrong place, read in the wrong place, if nothing else is there no problem.
With more pin read, this overflow is getting worse and start reading, or in very wrong places.

A quick fix would be to increase the sizes of these two stacks. But it might be cleaner to actually use the `adcPinToChannelNum` function already there for arduino to shift the pin number only, to use a stack of minimal size.

I'll first check if this is affecting all Teensys, and come with a fix using hopefully the second solution soon.
Let me know if you have any comments!

tomcombriat:
The last commit should - in theory - solves the ADC issues for all Teensys supported (3.X, 4.X).
The main problem with `mozziAnalogRead` on Teensy was that the pin number was used as index for the `analog_readings` array, even thought the smallest pin number was not 0. This led to overflows with more or less dramatic consequences. To fix that I made a map of the pins to give them a valid index in the expected range.

This has been tested on Teensy LC, soon to be on 3.2 and 4.0.

Happy to take comments if there is a more intelligent way to do this map :) !

Best,

tomcombriat:
Alright, the previous proposition solved the overflow in `analog_readings` but another problem occurred during testing. Some of these chips (basically all of them except the Teensy LC) have *two* internal ADC. The problem is that, each analogInput pin is not *necessary* connected to the two ADC, sometimes only to one. 

Henceforth, a test needs to be made when the chip has two ADC in order to know which one to use, the last commits 0c96d3 and so on are implementing that.
Tested on Teensy 3.2 and Teensy 4.0. Teensy LC should not have been affected by the last commits as it has only one ADC but will try to test it soon too…

Every time I try to modify "something simple" in Mozzi it ends up to be a big story, sorry for that.

tomcombriat:
Hello!

This is now also tested on Teensy LC with no problem. I also tried to update the Readme to reflect the latest changes. I was not very inspired so feel free to modify it! I put my name for the Teensy4 port, feel also free to remove it, this was mainly a modification of the original Teensy3 port, most of my commits were actually to fix other small related problems.

I think this is now in final state, or very close to it!
Best,

MozziGuts.cpp
```diff
@@ -395,13 +396,28 @@ void samd21AudioOutput() {
 #if !IS_AVR()
 static void startAudioStandard() {
 
+#if IS_TEENSY3() || IS_TEENSY4()
+  adc->adc0->setAveraging(0);
+  adc->adc0->setConversionSpeed(
+				ADC_CONVERSION_SPEED::MED_SPEED); // could be HIGH_SPEED, noisier
+  #ifdef ADC_DUAL_ADCS
+adc->adc1->setAveraging(0);
```
tfry-git, line 404: Indentation

tfry-git (commented):
Looks good to me save for that one indentation nitpick.

However, bonus points, if you can clean up the commit history some, i.e. squash the commits to maybe two or three (the port, the ADC fix, the documentation). It's not a strict requirement, but may help sorting out any unexpected issues, later (should there be any).

tomcombriat:
@tfry-git Everytime I learn new things… I just successfully squashed my commits into three. This was my first time ;).

I'll soon start a device using that port, if a minor bug escaped my and your vigilance it should be spotted fairly quickly.

Thanks for your help!

tfry-git:
Great!

So you're saying, we'll wait another short while before merging, so you can do some final testing. Right?

tomcombriat:
Hei,

Not really, I think this is mergeable state now, ADC problems for all Teensys are solved, alongside the slight out-of-tune of the output, plus the Teensy4 support. If something went underlooked it has to be fairly small as it went through tests. I was more meaning that these small details (if there are some) will probably be spotted in the future, by me or someone else (like the ADC issue), but this is basically working now.

PR #147, opened 2022-02-15
Fix compilation for STM32_arduino newest version

tomcombriat:
After reinstalling Arduino_STM32 to the latest version after a complete reinstallation of my computer, I could not compile Mozzi for STM32. Apparently, the function to attach the ADC interrupt changed. This fixes the compilation here (but I think this would need confirmation from @tfry-git ).

Best,

### DIFF
--- a/MozziGuts_impl_STM32.hpp
+++ b/MozziGuts_impl_STM32.hpp
@@ -46,7 +46,7 @@ void setupFastAnalogRead(int8_t speed) {
 }
 
 void setupMozziADC(int8_t speed) {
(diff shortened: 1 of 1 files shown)


tomcombriat:
This is for #145

tfry-git:
This seems to have changed, indeed (16 months ago; it's been a while since I updated...).

tomcombriat:
I am happy that this is still maintained, even it is getting hard to source good bluePills. Funny thing, I used to need to copy the ADC library for stm32 in the Mozzi folder before but this is no longer the case.

Issue #144, opened 2022-01-11
LowPassFilter coefficients incorrectly calculated


mzero:
The coefficients used in filter are incorrectly calculated.

The `fb` coefficient is calculated, effectively as:
```
fb = q + q * (1 - cutoff)
```
However, the proper equation, which is the comments at line 34, is:
```
fb = q + q / (1 - f)
```
The referenced algorithm at [musicdsp.org](https://www.musicdsp.org/en/latest/Filters/29-resonant-filter.html) correctly shows the division. The referenced fixed point code at dave's blog ([pawfal.org](http://www.pawfal.org/dave/blog/2011/09/betablocker-ds-test-tones/)) made the error of replacing the divide with a multiply. The error is noted in the comments to his post, and is now fixed the code in the post itself.

The effect of the error is that you can't get consistent resonance results with varying frequency.

---

The `f` coefficient is simply set to the input value `cutoff`, where `cutoff` maps the 8 bit fixed point value 0 to 255 to frequencies of 0 to ½ sample rate.

However, it turns out that _f_ in the DSP algorithm is computed as follows:
```
f = 2 * sin(pi * cutoff / samplerate)
```
The effect of setting f to the input parameter directly, is that it maps 0 to 255 to frequencies of 0 to 1/6 sample rate.

The approximation of _sin(x) = x_ for small _x_ is good enough for this purpose, but the range of the filter is much less than expected. Rather than being able to set the cutoff of this filter between 0 and 8kHz, you can only set it between 0 and 2,7kHz.

mzero:
The fix for `q` is straightforward, requiring a fix-point divide to be defined and used (and be sure to use `f`, not `cutoff` in the calculation).

The fix for `f` is depends on what the API should achieve:

1) Leave it as it is: The filter has a cutoff range of 0Hz to `AUDIO_RATE` / 2π - or about 2.6kHz for the 16384kHz sample rate, 5.2kHz with 32768kHz sample rate.

2) Push the available range by scaling the input by 1.5.
```
f = 1.5 * cutoff
```
This still makes use of the _sin x = x_ approximation, though pushes into less accurate territory. This raises the range to 3.9kHz  for the lower sampling rate, 7.8kHz for the higher.

3) The range is still rather low for a low pass filter - there is no way to set the center frequency high enough that it effectively goes away. You can solve this by increasing the range again:
```
f = 2.0 * cutoff
```
[...]

PR #152, opened 2022-05-08
More constant resonance over the frequency range for LowPassFilter

tomcombriat:
Hei,

This is a first response to #144 : the calculation explained in the last comment of that issue is applied here. 

Tests with high cutoff values (for instance 200 for the 8 bits version) of the initial filter indeed showed no resonance at all. This version keeps a more constant resonance over the whole range (slightly less at very high frequencies).

Note that this is not heavier on the MCU.

PS: I do not understand why is git showing three commits, with only one file changed… I tried to rebase my branch with no success… Sorry for that.


tfry-git:
Don't worry about the three commits, we can just squash merge them (incidentally, squash merging is also what started the trouble, as it has caused your master branch and sensorium:master to diverge; _after_ this is merged, use ```git reset --hard origin/master``` to get back in sync.)

As for the MR itself, I'll have to admit, I have no idea. Anybody else?

tomcombriat:
When analyzed carefully, the original version has a really non-constant resonance (from +14dB at low frequency to as less as 0.1dB at high frequencies). It diverges with the frequency approaching 0.

Unfortunately, the version proposed here has an overflow in `ucfxmul(q, SHIFTED_1 + cutoff);` : `SHIFTED_1 + cutoff` do not fit into the `su` type anymore, which also does not give a correct resonance. In order to have a more constant Q with the calculation explained I see only two solutions:

  - increase the container size for `SHIFTED_1 + cutoff` to the next type.
  - reduce both cutoff frequency and resonance parameters range to 0-127, allowing `SHIFTED_1 + cutoff` to fit in a `su` container holding, for the 8bits version, up to 255.

Any opinion?

tfry-git:
Yeah, some benchmarking is a good idea, although I believe your reasoning is sound, and there should not be any overhead.

tomcombriat:
In the end, I had to increase the cast in `ucfxmul` because, `SHIFTED_1 + cutoff` is now more than the size of `su` hence its multiplication with `q` (also of size `su`) potentially overflow the size of two `su`, especially at high resonance.

Hence, the multiplication is now happening to the next bigger size container, which I expect will affect the performances. How much I do not know yet, I have to benchmark it.

At least, the resonance as a function of the frequency is now flatter (not flat as the approximation becomes more and more invalid at high cutoff frequencies). As an example, a cutoff frequency of 200 now display a resonance of 4dB whereas it used to be less than a dB. The difference is quite hearable, especially at "high" frequencies: it sounds more like a resonant.

Depending on the results of the benchmark we can decide if it is worth or not. I'll keep it as draft for now.

tomcombriat:
Ok, I am fairly satisfied now.

My previous version (commit d6dd48d) was a bit slow. At the cost of one extra `byte` of memory I am now very close to original performances, with a more consistent Q.

## Benchmark for `setCutoffFreqAndResonance` (Arduino Mega, 8 bits version of the filter):

|LPF version | Timing (micro seconds) | Ratio with original|
|---|---|---|
|original| 4.92| 1|
|d6dd48d|7.68|1.56|
|last version| 5.48|1.12|

The drop in performance for the fast version is only caused by a binary shift. This version reduces a bit the accuracy of the Q correction with frequency, but stays very close to the "slow" version. Weirdly, these numbers are less than the comments in the original code… Or the compiler had become very smart (I had to force it to get these numbers) or the Mega is way more powerful than an Uno (I do not really believe so…).

## A few examples
### High frequency
#### Original (cutoff: 200, resonance 255)
old_200

#### This PR (cutoff: 200, resonance 255)
new_fast_200

The resonance is now present, even thought it is only +4dB

### Low frequency
#### Original (cutoff: 20, resonance 200)
old_20

#### This PR (cutoff: 20, resonance 200)
new_fast_20
 (the high frequency noise is due to saturation).

The two versions are equivalent at very low frequency, as the frequency increases the original version gets its resonance tamed out very fast. This version also shows a decrease in the resonance, because the approximation of $1/(1-f) \approx 1+1$ does not hold off, but this decrease is less pronounced, still showing a small resonance at high frequencies.

Tested for both versions of the filter (8 & 16 bits) on both 8 bitter (Arduino mega) and 32 bitters (Teensy 3.2)

tfry-git:
Looks good to me.

Could you insert a comment into the code explaining the idea / tradeoff behind the calculation?

tfry-git:
Afterthought: @sensorium , would it make sense to add @tomcombriat as "Collaborator"?

sensorium:
Great idea, Tom's certainly contributing a lot.

tomcombriat:
Hi,
Thanks very much @sensorium for the promotion and @tfry-git for suggesting it!
Considering that my PR usually get substantially improved before merging from discussion with you, be assured that I'll use push rights responsively and probably single-handedly only for small fixes.
Cheers,

PR #156, opened 2022-11-14
Delete Mozzi symlink

storrdev:
Arduino IDE Version: 2.0.1

I was receiving an error when attempting to install Mozzi according to the instructions in the README:

_In the Arduino IDE, navigate to Sketch > Include Library > Add .ZIP Library. At the top of the drop down list, select the option to "Add .ZIP Library"._

It seemed to be caused by the existence of the symlink `Mozzi` in the project root. Removing the symlink fixed the issue.

I have not contributed before, if there is a step I missed please let me know


sensorium:
Thanks Stephen, well spotted!

Issue #153, opened 2022-06-16
Wavefolder


cwmoriarty:
Has anyone considered developing a wavefolder for Mozzi?

See https://noiseengineering.us/blogs/loquelic-literitas-the-blog/getting-started-wavefolders

If there is interest, I could try my hand at creating one.

tomcombriat:
Hi,
I personally tried some time ago, but stopped when I wanted it not to be "spiky" upon folding, a bit more like an analog behavior…
You are of course welcome to try and submit something through a PR if you get some nice results!

cwmoriarty:
"spiky" with the sharp turn? Maybe a bit of slew would help?

If you have any of the old code, please send it along.

sensorium:
One (quick?) approach is to try Mozzi's WaveShaper class which uses a table to determine how the incoming wave gets shaped (or folded).  The demo is in the synthesis examples.  Scripts for producing tables are in Mozzi/extras/python (see table_generator_template.py), or you could draw tables in Audacity, save as raw 8bit and convert with char2Mozzi.py.

tomcombriat:
@cwmoriarty Yes, or just a LowPassFilter after the wavefolding. I have not trace of what I did, that was really a quick test. I think the part that can be tricky is to find, in an efficient way, how many times it has to be folded (if the folder is harsh enough so that it bounces back a few times). I am happy to help if you get started!

@sensorium Yes, that's an idea but if I understand correctly, using the WaveShaper prevents the levels at which the folding will occur to be changed at run-time, no? A big plus comes for waveshapper when levels can be bound to an envelope or so…

sensorium:
Yes, you're right Tom, using a single fixed table won't enable run-time changes in the mapping.
Just in case it's any use at all, I came across these quick sketches of wave folding and shaping in Pd (though it sounds like you're way ahead already):
folding: https://youtu.be/f2s3mAqLjkg
shaping: https://youtu.be/sg2TpxQ9CbU

PR #157, opened 2022-12-03
Wavefolder

tomcombriat:
Hello!

This is a response to #153: a simple wavefolder with settable high and low limits. The wave can be folded multiple times.
Not fully tested so far. Seems to work well on the scope.

Lacks at least an example before considering merging but if someone see an obvious optimization, I am all hear! I had to resort to use one division, which are notoriously slow on microcontrollers but I am not sure if the trick of computing the inverse on the changes of bounds only is actually worth it, open to comment!

Best,
Tom


tomcombriat:
PS: I do not understand why I have so many commits, I did a hard reset before creating this branch :(

PR #158, opened 2022-12-03
Wavefolder

tomcombriat:
Hello,

This is a simple wavefolder, in response to #153. Looks good on the scope but not fully tested yet. At least an example is missing.
Putting it as a draft for now, in case someone sees an obvious optimization!

I do not understand why there are so many commits, I tried to synchronise all my branch to upstream but they still show… I'm sorry for that.

Cheers,
Tom


tfry-git:
As for the wavefolding itself, I have no feedback. About the git-problem:

> I do not understand why there are so many commits, I tried to synchronise all my branch to upstream but they still show… I'm sorry for that.

Yeah, it's just far too easy to get into a mess with git...

I believe, the problem is that the master branch of your tomcombriat/Mozzi is not in sync with sensorium/Mozzi. For the future, you should try to avoid that situation, by doing _all_ your work in branches. For now, you are in a bit of a mess, as git thinks, your master branch has some changes that sensorium:master does not have. Actually they have been merged some time, but for git it still looks different.

Now, to get your _master_ branch back in sync, use:

1. Register sensorium/Mozzi as a separate "remote", so you can pull from it:  (you may have already done this, or something equivalent; ```remote show``` to list remotes):
```
remote add upstream git@github.com:sensorium/Mozzi.git
# or HTTPS-access: remote add upstream https://github.com/sensorium/Mozzi.git
```

2. Sync your master to upstream master:
```
git checkout master
git pull upstream master
git reset --hard upstream/master    # warning: deletes any local changes you may still have on master
git push origin master --force          #  warning: and now they are gone in tomcombriat/Mozzi:master, too
```

Now, after this, any new branch you create will be based on a state that git recognizes as something already in sensorium:master. However, your existing local branches are still messed up, and getting out of that may be non-trivial (```git rebase -i``` should work, in theory, or you can create a new "clean" work branches and ```git cherry-pick``` only the commits that are actually meaningful from the current branch). However, it may not be super-important to fix that, either, and, we can still squash commits while merging.

tomcombriat:
Hi,

Thanks for the git tips, I got most of it but I think I did a fetch instead of pull on the upstream.
Will try to do that and if it works might resubmit a cleaner PR (there is only one commit on this branch anyway).

Will add an example soon also.
Best,

tomcombriat:
Closing this in favour of the cleaner one I've just open (thanks @tfry-git , worked like a charm).

sensorium:
Nice one, Tom. 

I expect the / and % at audio rate will be trouble for slower processors!  I'll try to test it soon.

tomcombriat:
Yes, that was also my concern…

For the %, an easy optimization is to check only the least significant bit of the number. This is fairly common so I assumed the compiler might do it by itself. I can try to see if that makes a difference.

The division is a bit more tricky, but one way would be to compute the invert in the `setLimits` and use a multiply in `next()` will try.

tomcombriat:
Note that this has been superseeded by #159

PR #159, opened 2022-12-04
Added a simple wavefolder

tomcombriat:
Clean PR for the wavefolder, in response to #153 .

Will test it a bit more extensively and add an example soon.

### DIFF
--- a/WaveFolder.h
+++ b/WaveFolder.h
@@ -1,0 +1,108 @@
+/*
+ * Wavefolder.h
+ *
+ * Copyright 2022 Thomas Combriat
+ *
+ * This file is part of Mozzi.
+ *
+ * Mozzi is licensed under a Creative Commons
+ * Attribution-NonCommercial-ShareAlike 4.0 International License.
+ *
+ */
+
+#ifndef WAVEFOLDER_H
+#define WAVEFOLDER_H
+
+#include "AudioOutput.h"
+
+
+/* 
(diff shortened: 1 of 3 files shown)


tomcombriat:
Okay, at the cost of readability (have a look if you want ;) ), I gained a factor of 3 on the computation of the `next()` function.
Two optimizations:
  - `%` replaced by a bitwise and (I thought the compiler would have done it, but this version is slightly faster)
  - replace the hard division by a precomputation of the inverse. This will slow down the `setLimits` but accelerate (3x gain) the `next()` which (hopefully!) is computed more often. Had to use three more bytes for this. Also, this is a bit less precise for high numbers of large folds, because of the loss of precision in computing the inverse. This is minor though (error of 5 at most for a 10 times folded signal on a total range of 450…

Will do an example soon but open to comments in the meantime.
Best

tomcombriat:
Hello!
I think it starts to look good!
I have added an example, and tested on Arduino and Teensy.

For the last, there is a compilation problem as the compiler does not know with overload to choose between:
  - `static inline MonoOutput fromNBit(uint8_t bits, int16_t l)` and
  - `static inline MonoOutput fromNBit(uint8_t bits, int32_t l)`

A fix for that is coming soon.
Best,

tomcombriat:
Hi!
I am merging this as it is quite small and not impacting the main core. Will solve #153

PR #160, opened 2022-12-05
Fix compilation for Teensys

tomcombriat:
As said in #159 , Teensy would not compile on `static inline MonoOutput fromNBit` if the sample type is a plain `int` (which is actually the default audio type).

This fixes this problem but I'll wait for a go from @tfry-git to merge this one as I am not sure that multiplying the overloads is actually the good solution.

As for now (without this PR), the example in #159 does not compile on 32bits platforms, but works on AVR.

Best,
Tom

### DIFF
--- a/AudioOutput.h
+++ b/AudioOutput.h
@@ -106,6 +106,8 @@ struct StereoOutput {
   static inline StereoOutput fromNBit(uint8_t bits, int16_t l, int16_t r) { return StereoOutput(SCALE_AUDIO(l, bits), SCALE_AUDIO(r, bits)); }
   /** @see MonoOutput::fromNBit(), stereo variant, 32 bit overload */
   static inline StereoOutput fromNBit(uint8_t bits, int32_t l, int32_t r) { return StereoOutput(SCALE_AUDIO(l, bits), SCALE_AUDIO(r, bits)); }
+  /** @see MonoOutput::fromNBit(), stereo variant, native int bit overload */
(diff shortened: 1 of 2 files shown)


AudioOutput.h
```diff
@@ -106,6 +106,8 @@ struct StereoOutput {
   static inline StereoOutput fromNBit(uint8_t bits, int16_t l, int16_t r) { return StereoOutput(SCALE_AUDIO(l, bits), SCALE_AUDIO(r, bits)); }
   /** @see MonoOutput::fromNBit(), stereo variant, 32 bit overload */
   static inline StereoOutput fromNBit(uint8_t bits, int32_t l, int32_t r) { return StereoOutput(SCALE_AUDIO(l, bits), SCALE_AUDIO(r, bits)); }
+  /** @see MonoOutput::fromNBit(), stereo variant, native int bit overload */
+  static inline StereoOutput fromNBit(uint8_t bits, int l, int r) { return StereoOutput(SCALE_AUDIO(l, bits), SCALE_AUDIO(r, bits)); }
```
tfry-git, line 110: Could you try
```
template<typename T> static inline StereoOutput fromNBit(uint8_t bits, T l, T r) { return StereoOutput(SCALE_AUDIO(l, bits), SCALE_AUDIO(r, bits)); }
```
instead? If this works, it could replace both existing versions of the function, rather than adding a third.

tomcombriat: Will try.

tomcombriat: Seems to work. See latest commit.

PR #173, opened 2023-02-28
Compile and runtime error when EXTERNAL_AUDIO_OUTPUT is true

pschatzmann:
On an ESP32, when setting the EXTERNAL_AUDIO_OUTPUT to true I was running into compile errors.
After correcting them, I was running into runtime errors in my tests because the divider of 1 is not allowed.
After this correction the functionality is working again.

### DIFF
--- a/MozziGuts_impl_ESP32.hpp
+++ b/MozziGuts_impl_ESP32.hpp
@@ -104,11 +104,11 @@ static void startAudio() {
 #if (BYPASS_MOZZI_OUTPUT_BUFFER != true)  // for external audio output, set up a timer running a audio rate
   static intr_handle_t s_timer_handle;
   timer_config_t config = {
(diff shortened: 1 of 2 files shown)


tomcombriat:
Hi, 
Unfortunately I do not have the hardware to test this, hopefully someone else does around. Otherwise looks good!

pschatzmann:
I think it is pretty much low risk, because there is hardly any need for this. The output to I2S, the Internal DAC and PDM is basically all one needs. I just thought that it gives a bad picture to run into compile errors.

I tried to use this functionality, but it does not really fit into the audio architecture of the ESP32. I could just think of using this to output audio via PWM.

tomcombriat:
Alright, will merge very soon if the others have no objections.
Thanks!

PR #185, opened 2023-04-11
Added github CI workflow for Pico

tomcombriat:
I had a bit of problems to find the exact values, I might sound stupid but @tfry-git how do you reliably find the `fqbn` 0:)?

### DIFF
--- a/.github/workflows/compile_examples.yml
+++ b/.github/workflows/compile_examples.yml
@@ -45,6 +45,11 @@ jobs:
           - fqbn: arduino:samd:adafruit_circuitplayground_m0
             platforms: |
               - name: arduino:samd
+          - fqbn: rp2040:rp2040:rpipico
+            type: rp2040
+            platforms: |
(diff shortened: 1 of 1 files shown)


tomcombriat:
PS: I do not get why the reports are failing, are we asking too many things to be reported?

tfry-git:
> I had a bit of problems to find the exact values, I might sound stupid but @tfry-git how do you reliably find the fqbn 0:)?

I usually do not succeed on the first attempt, either :(. My current strategy is:
1. Look at the core in question, and copy from their workflow, if any.
2. Loot at platform.txt and boards.txt. In the latter, I believe the ".board" entries to contain the relevant id.
3. Trial and lots of error.

As for why the reports are failing, I have no real clue. However, the internets seem to think that intermittent problems are to be expected with github workflows, so I hope this will go away by itself (fingers crossed).

tomcombriat:
That's reassuring, somehow ;)

PR #189, opened 2023-04-29
Fixed compilation for EXTERNAL_AUDIO_OUTPUT for "new" STM32duino

tomcombriat:
Hi!
This fixes the compilation the EXTERNAL_AUDIO_OUTPUT feature for the new the STM port.

Note sure where is the best place to put this declaration though…


tfry-git:
Hm. Of course the define is meant to be configurable, so AudioConfigSTM32duino.h does look like the place for it - in theory.

In the old port, we seem to have cheated: https://github.com/sensorium/Mozzi/blob/b7350d2ccd4cfc56fdc0c4d2bde4b88f7df12bf6/MozziGuts_impl_STM32.hpp#L65

Maybe we should do the same, here, for now, and do a big sorting out of the config mess, another day?

PR #190, opened 2023-05-15
Creating a devel branch for Arduino R4

tomcombriat:
Hi!

This is a draft PR, just to have a discussion space for the port of the Arduino R4.

My initial attempt did not go very far, I cannot find the good variable for discriminating the board in the `hardware_defines.h`.

I am not completely sure, but I think there is a `{includes}` missing in the platform.txt file for this board. I would have expected something like `-DARDUINO_FSP {build.library_discovery_phase_flag}` in there. As it is not there this precomp variable is not set in the included library (and Mozzi for that extend). 

I am just wondering if I am on a completely wrong path here… If that sound legit, I'll ask Arduino about it.

### DIFF
--- a/CircularBuffer.h
+++ b/CircularBuffer.h
@@ -5,6 +5,10 @@ On 18 April 2014, the simplified version on the Wikipedia page for power of 2 si
 doesn't work - cbIsEmpty() returns true whether the buffer is full or empty.
 */
 
+#define MOZZI_BUFFER_SIZE 256 // do not expect to change and it to work.
+                              // just here for forward compatibility if one day
+                              // the buffer size might be editable
+
 /** Circular buffer object.  Has a fixed number of cells, set to 256.
 @tparam ITEM_TYPE the kind of data to store, eg. int, int8_t etc.
 */
@@ -47,9 +51,13 @@ public:
 	unsigned long count() {
 		return (num_buffers_read << 8) + start;
 	}
+        inline
+	ITEM_TYPE * address() {
+	  return items;
+	}
 
 private:
-	ITEM_TYPE items[256];
+	ITEM_TYPE items[MOZZI_BUFFER_SIZE];
 	uint8_t         start;  /* index of oldest itement              */
 	uint8_t         end;    /* index at which to write new itement  */
 	uint8_t         s_msb;
--- a/MozziGuts_impl_ARDUINOFSP.hpp
+++ b/MozziGuts_impl_ARDUINOFSP.hpp
@@ -63,22 +63,26 @@ void stm32_adc_eoc_handler() {
 
 
 //// BEGIN AUDIO OUTPUT code ///////
-FspTimer timer1;
+
 
 #if (EXTERNAL_AUDIO_OUTPUT != true) // otherwise, the last stage - audioOutput() - will be provided by the user
 inline void audioOutput(const AudioOutput f) {
+  /*
   analogWrite(AUDIO_CHANNEL_1_PIN, f.l()+AUDIO_BIAS);
 #if (AUDIO_CHANNELS > 1)
   analogWrite(AUDIO_CHANNEL_2_PIN, f.r()+AUDIO_BIAS);
-#endif
+  #endif*/
 }
-#endif
-
+#else  // manually setting a timer for EXTERNAL_AUDIO_OUTPUT
+FspTimer timer1;
 void timer_callbar_dummy(timer_callback_args_t __attribute__((unused)) *args){defaultAudioOutput();};
+#endif
 
 static void startAudio() {
 
-  pinMode(AUDIO_CHANNEL_1_PIN,OUTPUT);
+  //pinMode(AUDIO_CHANNEL_1_PIN,OUTPUT);
+
+#if (EXTERNAL_AUDIO_OUTPUT == true)
   timer1.set_period(1.0 / AUDIO_RATE);
   timer1.set_irq_callback(timer_callbar_dummy);
 
@@ -86,10 +90,8 @@ static void startAudio() {
   uint8_t type=0;
   int8_t tindex = FspTimer::get_available_timer(type);
 
-    if (tindex < 0) {
-    tindex = FspTimer::get_available_timer(type, true);
-  }
-
+  if (tindex < 0) tindex = FspTimer::get_available_timer(type, true);
+    
   if (tindex >= 0) {
     // defaultAudioOutput cannot be used straight, so we are using a dummy, with no argument, linking to defaultAudioOutput
     timer1.begin(TIMER_MODE_PERIODIC, type, tindex,AUDIO_RATE, 50.0f, timer_callbar_dummy);
@@ -97,6 +99,7 @@ static void startAudio() {
     timer1.open();
     timer1.start();
   }
+#else
 
   
   //Serial.println("sa");
@@ -106,12 +109,15 @@ static void startAudio() {
   // 1) setting up some DAC mechanism (e.g. setting up a PWM pin with appropriate resolution
   // 2a) setting up a timer to call defaultAudioOutput() at AUDIO_RATE
   // OR 2b) setting up a buffered output queue such as I2S (see ESP32 / ESP8266 for examples for this setup)
-#if (EXTERNAL_AUDIO_OUTPUT != true)
   // remember that the user may configure EXTERNAL_AUDIO_OUTPUT, in which case, you'll want to provide step 2a), and only that.
   #endif
 }
 
 void stopMozzi() {
+#if (EXTERNAL_AUDIO_OUTPUT)
   timer1.stop();
+#else
+
+#endif
 }
 //// END AUDIO OUTPUT code ///////
--- a/MozziGuts_impl_ARDUINOFSP.hpp
+++ b/MozziGuts_impl_ARDUINOFSP.hpp
@@ -14,6 +14,8 @@
 #  error "Wrong implementation included for this platform"
 #endif
 
+#include <FspTimer.h>
+
 
 ////// BEGIN analog input code ////////
 
@@ -61,6 +63,7 @@ void stm32_adc_eoc_handler() {
 
 
 //// BEGIN AUDIO OUTPUT code ///////
+FspTimer timer1;
 
 #if (EXTERNAL_AUDIO_OUTPUT != true) // otherwise, the last stage - audioOutput() - will be provided by the user
 inline void audioOutput(const AudioOutput f) {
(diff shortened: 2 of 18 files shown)


tomcombriat:
Ok found it. Unfortunately I will have not have time to work on that for the remaining on the week…

tomcombriat:
(Just to precise, as I never enough time to go very far, I commit things I am doing on this on the go, even thought a big step might not be finished, I think we will probably need to squash some of these commits in the end!).

Compiles now, but nothing implemented.

tomcombriat:
Alright, things got a bit harder than expected, but at least a first version is working!
I am converting this PR to "ready" even though it should not be merged yet, more to indicate that I think that is now room for eventual discussion. Furthermore a bit of cleanup, squashing and documentation will likely be welcome.

## Some details

### Output
There is a single DAC pin (A0) on this board, the configuration is using that by default.
#### Pros
 - cleaner sound 
 - can work straight from a buffer
#### Cons
- HiFi/Stereo cannot work as there is only one such pin…
- specific to *that* pin only.


### Under the hook
As said this DAC can work straight from a buffer (no need to commit the buffer for instance), and I chose to use a Mozzi circular buffer for that that is actuated along the way and the DAC just reads from that. Unfortunately, I had to add a access to the data that the buffer contains, I think it is not very problematic but there might be some debates here.

The DAC sample frequency can be set by a timer (you can have a look at the code, it is not completely straightforward) which is quite neat but unfortunately I did not find any way to know how many samples has been played. Hence, as the DAC is not reading the samples from the buffer *per se*, one needs to tell the buffer that some samples can be removed in order to make room for the next ones. 

And I had to use *another* timer for that… I could not find a way to use the same timer for both updating the DAC and calling and IRQ to remove the samples from the buffer (setting up the IRQ stops the DAC or vice-versa). There might be a trick somewhere but I could not find it. 

#### Others

- EXTERNAL_AUDIO_OUTPUT should work but untested
- asynchroneous `analogRead`s are not implemented (even thought, from what I have seen it should not be too hard (1))
- only the DAC output is implemented but others like PDM should not be very hard (1)

(1): I should be more careful when saying that, it always ends up being a lot more work than expected ;).

tomcombriat:
Hello!

This is now tested (and working) with the Uno R4 WiFi!

I would be in favor of merging that, even though the `mozziAnalogRead()` is not implemented yet and maybe rolling out a version when this is implemented but as I made a few changes deep in Mozzi I would prefer to have a second opinion!

MozziGuts_impl_ARDUINOFSP.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+#if EXTERNAL_AUDIO_OUTPUT == true
+void timer_callback_dummy(timer_callback_args_t __attribute__((unused)) *args){defaultAudioOutput();};
+#else
+//void timer_callback_dummy(timer_callback_args_t __attribute__((unused)) *args){
+void timer_callback_dummy(timer_callback_args_t __attribute__((unused)) *args){output_buffer.read();}; // to empty the buffer (the dac does not take care of it), a bit a waste of timer...
+#endif
+
+void timer_init() {
+  uint8_t type = 0;
```
tfry-git, line 166: What is type 0? Please add a comment.

MozziGuts_impl_ARDUINOFSP.hpp
```diff
@@ -0,0 +1,227 @@
...
+#if EXTERNAL_AUDIO_OUTPUT != true
+  dac_creation(AUDIO_CHANNEL_1_PIN);
+#endif
+  timer_init();
+#if EXTERNAL_AUDIO_OUTPUT != true
+  dac_init();
+  R_DTC_Open(&dtc_ctrl, &dtc_cfg);
+  R_DTC_Enable(&dtc_ctrl);
+
+  dtc_cfg.p_info->p_src = output_buffer.address();
```
tfry-git, line 215: So, do I get this right? The DAC will keep looping over the buffer at AUDIO_RATE. Further down, below, you are advancing the read position as known to Mozzi, also at AUDIO_RATE.

Nifty, but please add a short explanation.

MozziGuts_impl_ARDUINOFSP.hpp
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  if (tindex >= 0) {
+    timer.begin(TIMER_MODE_PERIODIC, type, tindex, AUDIO_RATE, 50.0,timer_callback_dummy);
+    timer.setup_overflow_irq();
+  }
+  
+#if EXTERNAL_AUDIO_OUTPUT != true // we need to set up another timer for dac caring
+  // note: it is running at the same speed than the other one, but could not manage
+  // to get the other one updating the dac and removing the samples from the buffer…
```
tfry-git, line 180: Why not? If you simply add timer_overflow_dummy as a callback timer_dac, shouldn't that be just enough?

tfry-git (commented):
Nice! The usual nitpick, and some questions in individual comments.

Note that I have not done _any_ testing, myself.

CircularBuffer.h
```diff
@@ -5,6 +5,10 @@ On 18 April 2014, the simplified version on the Wikipedia page for power of 2 si
 doesn't work - cbIsEmpty() returns true whether the buffer is full or empty.
 */
 
+#define MOZZI_BUFFER_SIZE 256 // do not expect to change and it to work.
```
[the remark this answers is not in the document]
tomcombriat: I did that exactly because one of my long term idea would be to make the mozzi buffers adjustable in sizes. As I think this is the only thing that would require to be adapted in Mozzi for different buffer sizes (except the buffer themselves of course), I was anticipating a bit and not wanted to bury that to far down in the code.

MozziGuts_impl_ARDUINOFSP.hpp
```diff
@@ -0,0 +1,227 @@
...
+volatile uint32_t pin;
+CircularBuffer<uint16_t> output_buffer;
+uint8_t dac_bits;
+dtc_instance_ctrl_t dtc_ctrl;
+transfer_info_t dtc_info;
+dtc_extended_cfg_t dtc_cfg_extend;
+transfer_cfg_t dtc_cfg;
+
+////////////// DAC CREATION AND FIRST INIT ///////////////////
+void dac_creation(pin_size_t pinN) {
```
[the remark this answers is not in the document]
tomcombriat: I agree, but I have to say now that I do not remember how verbatim it is from the source. Will try to come up with a more modular solution.

MozziGuts_impl_ARDUINOFSP.hpp
tomcombriat: Good question, that's part of something I copied from somewhere so do not really know, will try to find out.

MozziGuts_impl_ARDUINOFSP.hpp
tomcombriat: Yeah, after searching for quite some time the best way to interact with this DAC I found out that seems to be the more satisfying (updating the buffers as a bunch is way to slow). Will add some explanation as this is deviating quite a lot from the previous strategies used in Mozzi.

MozziGuts_impl_ARDUINOFSP.hpp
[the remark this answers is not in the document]
tomcombriat: Done

MozziGuts_impl_ARDUINOFSP.hpp
```diff
@@ -0,0 +1,227 @@
...
+#ifdef DAC_ADDRESS_8_CH0
+        dtc_info.p_dest = (void *)DAC_ADDRESS_8_CH0;
+#endif
+      } else if (GET_CHANNEL(pin) == 1) {
+#ifdef DAC_ADDRESS_8_CH1
+        dtc_info.p_dest = (void *)DAC_ADDRESS_8_CH1;
+#endif
+      }
+    } else {
+      dac_bits = 12;
```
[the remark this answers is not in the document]
tomcombriat: Actually, this is drawn from the DAC capacity. I think it would be smarted to state that this port is 12bits, and 12 bits only (this function gets it from the pin it uses to output Audio, not sure how configurable it is).

MozziGuts_impl_ARDUINOFSP.hpp
tomcombriat: This is actually just an initialization: `type` is updated by `tindex = FspTimer::get_available_timer(type, true);` once a Timer get selected. I removed the default value, which is not bringing anything except confusion ;).

MozziGuts_impl_ARDUINOFSP.hpp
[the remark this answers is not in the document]
tomcombriat: I kept the timer init in the initial file as it is really bounded to Mozzi (AUDIO_RATE and so on). Also, this is doubtful to change I hope.

MozziGuts_impl_ARDUINOFSP.hpp
tomcombriat: I left this one for when I'll have a lot of time in front of me (or someone else solves it, if possible). I remember trying quite hard to make this work to not avail and getting back into the not so commented source code of these timers always takes a bit of time…

tomcombriat: Added some documentation about the strategy used here as it is fairly different from before.

MozziGuts_impl_ARDUINOFSP.hpp
tomcombriat: I left this one for either when I have a good chunk of uninterrupted (pun intended) time before me, or someone else. I remember trying quite hard to make that work with one timer, to not avail. Getting back into the more or less document source files of these timers gets some time.

tomcombriat: (From what I remember, adding the overflow_dummy as a callback_timer_dac would mess up with the DAC, which would not get updated anymore)

tomcombriat:
Alright, I took into account most of your comments @tfry-git . Let me know what you think.

You will see that I left out trying to get everything on one timer: I remember banging my head on it before but I cannot afford to spend the night on it today ;). I honestly think that this go in like that for now, as it is working but with acknowledging that:
 - the async ADC is not implemented yet
 - it *might* be possible to reduce to one timer

Cheers,

tfry-git (approved):
Note: I still have not done any testing, myself. Time is flying by...

Looks good enough to me, code-wise, and we don't have to be perfect & complete right from the start.

One thing I've still been thinking a little about is the usage of CircularBuffer. I think your solution is really spot on, in what it does, but perhaps it would be cleaner after all to do the same by hand on a sample buffer? At least one thing that is going to fail is any extension to STEREO (well, for now, there is only one DAC channel, so...).

But again, good enough to go in, for now.

tomcombriat:
@tfry-git  Ok thanks, will merge that then! Next target will be async ADC for this platform probably.

I have to say I did not fully understand your last paragraph. Do you mean to make manually a circular buffer working on the "native" samples for the Renesas? Or the native ones of Mozzi? The way I saw this is, if a platform of this arch with two DACs should arise, then the modification to this code should be minimal (adding a other buffer and branching it, it can be advanced by the same timer than the first one).

On the wider way of thinking, apart from making the buffers customizable in size, maybe we should start to think in terms of a pool of buffers? For instance creating a pool of one or two buffers for mono and stereo and working on that (instead of reading them separately) which would potentially remove a good chunk of preproc for the stereo case? Of course that might also make the way easier for more channels… These are just thoughts, not sure how good it would turn in practice though.

PR #192, opened 2023-05-16
Fix randSeed()

### LINKED ISSUE #191: The randSeed() argument type is different in the header and source files.
Hello, I am trying to use randSeed() and I am getting a link error.

When I tried to use randSeed() I was getting a link error.
The reason is that the type of randSeed is different in mozzi_rand.h and mozzi_rand.cpp. xorshiftSeed was also the same.

**mozzi_rand.h**

- void randSeed(unsigned long seed);
- void xorshiftSeed(unsigned long seed);

**mozzi_rand.cpp**

- void randSeed(long seed);
- void xorshiftSeed(long seed);

I am grateful for your projects.
Thanks in advance.

tomcombriat:
Should fix #191 .

Hopefully I picked the correct one (looked a bit briefly).

### DIFF
--- a/mozzi_rand.cpp
+++ b/mozzi_rand.cpp
@@ -50,7 +50,7 @@ for randomSeed(). randSeed is the same as xorshift96Seed(), but easier to
 remember. 
 @param seed a number to use as a seed.
 */
-void randSeed(long seed)
+void randSeed(unsigned long seed)
 {
 	x=seed;
 }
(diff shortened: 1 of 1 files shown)


sensorium:
Ah, sorry I missed this, Tom.  I just made the change myself.  Being unusually diligent while procrastinating on something else...

tomcombriat:
@sensorium Absolutely no problem! The goal is that to be fixed, whatever how :)!

Closing this.

PR #199, opened 2023-09-11
Asynchroneous ADC (mozziAnalogRead) for Uno R4

tomcombriat:
Hi!

This is a proposal for implementing `mozziAnalogRead()` on the Uno R4, following #190.
The result is a bit of a jigsaw puzzle between different source code I could find online (from Arduino, and from Renesas)… There are not a lot of documentation on this chip yet, not to say well documented for people green behind the ears like me…

Anyway, this seems to work quite well. I put the sources inside a separate file because they are not fancy to read but when I look at the final result it might be able to go inside the implementation file. I am up for opinions on this but also on the code in general. I might be doing things a bit naively in there but I had a bit of troubles to make that works in the first place.

I think that when this gets in there it might be time for a new release for the two newer ports (giga and R4). I do not mind trying to do it!

Cheers,

### DIFF
--- a/MozziGuts_impl_RENESAS.hpp
+++ b/MozziGuts_impl_RENESAS.hpp
@@ -20,11 +20,11 @@
 ////// BEGIN analog input code ////////
 
 
-#define channelNumToIndex(channel) channel-14
-void const *const p_context = 0;
-adc_callback_args_t *const p_callback_memory = NULL;
-uint16_t cfg_adc = 0;
-uint8_t r4_pin;
+#define channelNumToIndex(channel) channel-14  // A0=14
+void const *const p_context = 0; // unused but needed for the ADC call
+adc_callback_args_t *const p_callback_memory = NULL;  // idem
+uint16_t cfg_adc = 0;  // put at 0 for starters but modified by the ADC_container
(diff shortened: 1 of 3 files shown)


tomcombriat:
Hi!

If there is no objections I will merge that soon and try to bundle it into a new release. Feedback might come from users also!

Best,
Tom

sensorium:
Hey Tom,
I don't have the knowledge to review this, but if you're ready to face the blowtorch of user feedback, I have no objections...

sensorium:
Oh, do the docs need re-generating to include your changes?  I can do that in the next couple of days if needed.

tomcombriat:
Hi @sensorium , yes, I am ready for the blowtorch, I think putting it in the frontline is sometimes the best way to discover the potential bugs.

I don't think the doc needs re-generating as the API basically did not change.
Will do that soon then, thanks!

tomcombriat:
re-Tested on both Wifi and Minima and seems to work quite well. Merging.

tomcombriat:
@sensorium I prepared a draft new version that I think could be released. Will wait a few days in case you want to add something! Feel free to modify it or to redo it!

PR #202, opened 2023-10-11
Fix frequency offset for AVR

### LINKED ISSUE #201: Output frequency 0.25% off on AVRs
Hi,
Just reporting here something that has been said on the [forum](https://groups.google.com/g/mozzi-users/c/ynNL_x_yoRs): when playing a simple oscillator at a given frequency on an AVR board, the output is slightly off, to be precise it is too low by 0.25% from my tests.

This is not a lot, but it is hearable when playing when other very in-tune digital instruments. 

Looking at the timer for AVR source code:

`  Timer1.initializeCPUCycles( F_CPU / AUDIO_RATE,`

should only be off by (16e6/32768 - int(16e6/32768))/32768 = 0.0576%, five times less!

So if anyone has an idea!

tomcombriat:
Hi,

Fixes #201 by fine tuning the timer frequency. Basically, we get closer to the "ideal" timer frequency by this. Output frequency detuning shifts from 0.25% to 0.06%. To test this, I generated sinusoidal tones in Audacity and listen to the beating when superimposing them to the same tone generated by Mozzi. HiFi is 0.06% off in the native version so I did not touch that.

Untested with external audio, but that should behave the same. Of course, this changes the behavior of all sketches on AVR (but to be closer to reality which I think is fine!) 

Would love to have confirmation from someone else owning an AVR to be sure I am not looking to fix problems with the particular board I am testing with!

### DIFF
--- a/MozziGuts_impl_AVR.hpp
+++ b/MozziGuts_impl_AVR.hpp
@@ -168,7 +168,7 @@ static uint8_t mozzi_TCCR4A, mozzi_TCCR4B, mozzi_TCCR4C, mozzi_TCCR4D,
 static void startAudio() {
   backupPreMozziTimer1();
   Timer1.initializeCPUCycles(
-      F_CPU / AUDIO_RATE,
+      F_CPU / AUDIO_RATE-1,
       PHASE_FREQ_CORRECT); // set period, phase and frequency correct
   TIMSK1 = _BV(TOIE1); // Overflow Interrupt Enable (when not using
                        // Timer1.attachInterrupt())
@@ -198,10 +198,10 @@ static void startAudio() {
   //	pinMode(AUDIO_CHANNEL_2_PIN, OUTPUT);	// set pin to output for audio
 #  if (AUDIO_MODE == STANDARD)
   Timer1.initializeCPUCycles(
-      F_CPU / AUDIO_RATE,
+      F_CPU / AUDIO_RATE-1,
       PHASE_FREQ_CORRECT); // set period, phase and frequency correct
 #  else // (AUDIO_MODE == STANDARD_PLUS)
-  Timer1.initializeCPUCycles(F_CPU / PWM_RATE,
+  Timer1.initializeCPUCycles(F_CPU / PWM_RATE-1,
                              FAST); // fast mode enables higher PWM rate
 #  endif
   Timer1.pwm(AUDIO_CHANNEL_1_PIN,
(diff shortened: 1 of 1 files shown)


tfry-git:
I have _not_ done any testing. I suppose this makes some sense, though: Timer1.initializeCPUCycles() actually just seems to set the timer compare register to that value. So if I understand that correct (which I am not sure about), a value of 0 means to trigger every timer tick, 1 every other timer tick, and so on. 

What I find a bit unsettling, is that the formulas in AVRs own application note do not include the -1.

At any rate, for better readability, I suggest adding braces as ```(F_CPU/AUDIO_RATE)-1``` (or at least use even spacing around "/" and "-"). 

--- Reference:

https://ww1.microchip.com/downloads/en/Appnotes/Atmel-2505-Setup-and-Use-of-AVR-Timers_ApplicationNote_AVR130.pdf

tomcombriat:
Hi,

Thanks @tfry-git !

> At any rate, for better readability, I suggest adding braces as (F_CPU/AUDIO_RATE)-1 (or at least use even spacing around "/" and "-").

Done.

For the rate, I do not really understand why the -1 is needed, but maybe the timer is losing one clock cycle somewhere? In any case, it seems to be quite closer to the ideal case (there is always a rounding), but I just want to be sure it is not a local defect on the board I am testing on.

tomcombriat:
Any comment/test for this?

I have drafted a new release which could (or not) include this. More importantly it is adding two new boards (Giga and Uno R4). It is one click away from being published, if you do not have concerns I'll do it in a few days!

tfry-git:
My first approach was to check what the CPU itself has to say about it: I added debug code inside the timer 1 ISR to measure the time spend for 10*AUDIO_RATE samples (based on micros()).

This yields 9994212 micros (with the patch) vs. 10014696 micros (without). Those numbers are fairly stable (with an occasional deviation of 4 micros), and are identical at AUDIO_RATE 16384 and 32768. -> Slightly better with the patch.

I then used a simple updateAudio like this:
```AudioOutput_t updateAudio(){
  if ((audioTicks() / 256) %2) return 244;
  return -243;
  //return MonoOutput::from8Bit(aSin.next()); // return an int signal centred around 0
}```

For a scope-friendly rectangle wave at theoretically 32 Hz. On my crap-scope that yielded a stable 32.02 Hz with the patch and 31.95 Hz without. -> Again slightly better with the patch. Results were again remarkably similar for AUDIO_RATE 16384 or 32768 (adjusting the divisor, of course).

Testing done on a low quality Uno clone, but in particular the experiment using on-board timing with micros() seems to provide fairly solid backing that this is indeed the correct thing to do. Tested STANDARD_PLUS, only.

Please do add a comment pointing to the issue / MR near that ```-1```.

tomcombriat:
Ok, great, thanks for testing!

I'll add some comments to these tests and merge soon then.

tomcombriat:
Added some comments and again made a mess with git trying to squash my last two commits… Sorry for that. Will merge once tests are completed.

PR #203, opened 2023-10-21
Github workflow: Revert version increas in checkout action

tfry-git:
To test, whether this fixes the "report" stage on PR, or whether that breakage is due to something else.


tfry-git:
Yup, seems to work, where version 4 somehow didn't. No idea, why, but let's go back, for now.

(See https://github.com/sensorium/Mozzi/pull/198)

tomcombriat:
Indeed, sorry!

Issue #205, opened 2023-10-22
RFC: Mozzi 2.0 part 2 - Configuration restructuring


tfry-git:
As a continuation of #204 but not strictly dependent, here are some thoughts on what I would like to change regarding the configuration itself.

The first change I propose is easy enough to summarize: Prefix each Mozzi configuration ```#define``` with ```MOZZI_``` in order to avoid name clashes. Since this only affects customization of the configuration, the extra typing should not be too cumbersome. Exceptions will be made for AUDIO_RATE and CONTROL_RATE, which are also commonly used inside existing sketches.

The second proposed change is the general setup for configuration. This would consist of:

  - MozziCustomConfig.h This file will simply contain definitions to be used in the configuration, i.e. (and I'm actually trying to be complete, here):
```
#define MOZZI_MONO 1
#define MOZZI_STEREO 2

#define MOZZI_OUTPUT_PWM 1  // Corresponds to "STANDARD_PLUS" on AVR; "STANDARD" has been deprecated for ages, let's drop it
#define MOZZI_OUTPUT_2PIN_PWM 2  // Formerly known as "HIFI"
#define MOZZI_OUTPUT_EXTERNAL 3  // Formerly known as EXTERNAL_AUDIO_OUTPUT
#define MOZZI_OUTPUT_PDM_VIA_I2S 4
#define MOZZI_OUTPUT_PDM_VIA_SERIAL 5
#define MOZZI_OUTPUT_I2S_DAC 6
#define MOZZI_OUTPUT_INTERNAL_DAC 7

#include <hardware_defines.h>
```

I.e. we're actually trying to use exactly the same config values across all boards. Naturally, not each value will be supported on each board.

Next, the user may (or may not) customize some of these options like so:
```
#include <MozziCustomConfig.>
#define AUDIO_RATE 32768
#define MOZZI_AUDIO_MODE MOZZI_OUTPUT_EXTERNAL
// [...]

#include <MozziGuts.h>
```

Finally, MozziGuts.h will (automatically) apply defaults, where nothing has been set, and check for invalid settings. This will likely be mostly handled inside (internal) platform specific files like, e.g.
  - mozzi_check_config_avr.h

```
#if not defined(AUDIO_RATE)
#define AUDIO_RATE 16384
#endif

#if not defined(MOZZI_USE_AUDIO_INPUT)
#define MOZZI_USE_AUDIO_INPUT false
#endif

#if not defined(MOZZI_AUDIO_MODE)
[...]

Issue #204, opened 2023-10-21
RFC: Mozzi 2.0 part 1: Single compilation unit


tfry-git:
This is a write-up of something I have been pondering for quite some time, now: One of the larger problems with Mozzi is that in order to change something in the config (and the need to do so is not uncommon at all), you have to modify config files in the library itself. To spell out some of the problems with that:

  - It's really unfriendly to new users
  - Dependent projects (using Mozzi) that require a certain config have to pass the same problem to their users
  - It's really annoying when working on several projects in parallel
  - It prevents us from having automated testing coverage on non-default configuration options such as stereo output

That problem has been felt, before, so why didn't we fix it, yet?
  - Switching to run-time configuration is not an option (or at least not for everything), for performance and code-size concerns
  - We cannot currently allow users to e.g. customize ```AUDIO_RATE``` at the top of their sketch

Why can't we have the latter? Compilation units. For a very rough summary, every .cpp-file is compiled, separately, and only in a final stage everything is linked together. One such .cpp-file is the user's sketch. Another, for example is MozziGuts.cpp. Both also include several common .h-files, but importantly, where they make use of defines such as ```AUDIO_RATE```, this has to match in all compilation units. If ```AUDIO_RATE``` was defined at the top of the user-sketch, the MozziGuts.cpp compilation unit would not "see" that define. The current way to make things work, is thus to have both MozziGuts.cpp and the user sketch include a common configuration file. Due to the way the arduino build system works (I won't go into details, mostly, because I don't have them at the top of my head, but I _did_ investigate, before), this common configuration file can only reside in the library folder itself. -> The situation we are in.

So what's the alternative? Putting everything into a single compilation unit. Note this does not mean a single code file, but rather than having MozziGuts.cpp include MozziGuts.h, we'd have MozziGuts.h include (a renamed) MozziGuts.hpp (and the same for all other .cpp files inside Mozzi). In effect, starting with the user sketch as the first file the preprocessor will lump all ```#include```d code files together into one big file, and compile that in a single go. Thus a ```#define``` near the top of the user sketch will affect all Mozzi code files, too.

What are the drawbacks of that plan?
  - Compilation times may go up a bit, because any change in the user sketch will require a recompilation of everything inside Mozzi. I don't really see this as an issue, though. Mozzi code size is not that huge in the first place. The majority of compilation times "heavy" platforms such as ESP32 is spent in other libraries that will not be affected.
[...]

PR #211, opened 2023-11-13
Mozzi 2.0 - Start config rework

tfry-git:
See #205 .

Status: Basically complete.

Some tasks not to forget:
- [ ] Check for TODO comments left over in the touched code
- [x] Clean up the old AudioConfigXYZ files and mozzi_config.h
- [x] Check for references to mozzi_config.h in API docs and code comments
- [x] Testing (automated testing coverage has to wait for #204 )

Testing matrix - for most boards I intend to test the default mode, only. Later, additional configs will at least be compile-tested outmatically (depends on the single compilation unit work).
| Platform & feature       | Compiles     | Works     |
|----------------------------------|------------------|--------------|
| AVR - default                  | yes                | yes          |
| AVR - 2 pin (HIFI)           | yes                | yes          |
| AVR - stereo                   | yes                | yes          |
| AVR - external                |  yes                   |   yes           |
|----------------------------------|------------------|--------------|
| ESP32 - default                |  yes                    |  yes            |
|----------------------------------|------------------|--------------|
| ESP8266 - default                |  yes                    | yes             |
|----------------------------------|------------------|--------------|
| MBed (Giga) - default                | yes                     | yes       |
| MBed (Giga) - external                | yes                     | yes       |
|----------------------------------|------------------|--------------|
| UNO R4 - default                | yes                     |  yes       |
| UNO R4 - external                | yes                     | yes  |
|----------------------------------|------------------|--------------|
| RP2040 - default                | yes                     | yes             |
| RP2040 - I2S LSBJ                | yes                     | yes             |
| RP2040 - external                | yes                     | yes             |
|----------------------------------|------------------|--------------|
| SAMD21 - default                | yes                    |              |
|----------------------------------|------------------|--------------|
| STM32duino - default                |  yes                    | yes             |
| STM32duino - external                |  yes                    | yes             |
|----------------------------------|------------------|--------------|
| STM32maple - default                |  yes                    | yes             |
| STM32maple - external                |  yes                    | yes             |
|----------------------------------|------------------|--------------|
| Teensy3.2 - default                |  yes                    | yes             |
| Teensy3.2 - external                |  yes                    | yes             |
| TeensyLC - default                |  yes                    | yes             |
| TeensyLC - external                |  yes                    | yes             |
|----------------------------------|------------------|--------------|
| Teensy4 - default                |   yes                   | yes             |
| Teensy4 - external                |   yes                   | yes             |
|----------------------------------|------------------|--------------|


tfry-git:
Ok, there are rough edges left (regarding documentation), and testing is still needed, but the basic shape is complete.

Sorry that it turned out to be such a huge diff.

Some notes:
  - As I had to do a lot of documenting, anyway, I found it useful to de-duplicate some of this by moving hardware-specific notes out of the primary readme. Reduces the chances of documentation getting out-of-sync with the implementation, I hope.
  - I tried not to actually change the implementations, but at some points I did. For instance most ports used to - unnecessarily - set up a timer at audio rate, even in external output with bypassed buffer (now called "MOZZI_OUTPUT_EXTERNAL_CUSTOM"). Others failed to implement MOZZI_OUTPUT_EXTERNAL_TIMED, although it's trivial to add.
  - While at it, I added a config option to disable analog input code. See documentation for MOZZI_ANALOG_READ-define.

tomcombriat:
Alright, will try to have a look at it soon. Will start with having a look to try to understand what has changed and will then complete your testing matrix, also with `EXTERNAL_AUDIO`. I see that you managed to save up some RAM in the process!

> For instance most ports used to - unnecessarily - set up a timer at audio rate, even in external output with bypassed buffer

Which are these exactly? They are not of platforms which work on an external_buffer if I reckon correctly (RP2040 and ESP?)

Allow a bit of time for the testing ;)

PS: if you have a bit of postpartum depression, the FixMath2 will probably be ready for your comments soon ;)

tfry-git:
> I see that you managed to save up some RAM in the process!

I believe, that's the analog read code (now more properly disabled on platforms that don't implement it).

> Which are these exactly? They are not of platforms which work on an external_buffer if I reckon correctly (RP2040 and ESP?)

Frankly, I didn't keep track. Might have included (some of) those, too, though, which is not all that surprising, since the logic was a bit difficult to parse (external being orthogonal to the output mode, and BYPASS_MOZZI_OUTPUT_BUFFER again independent from that). Some ports remembered to set up an audio rate timer when external output was set (and others forgot), but did not check for bypassed buffer, there. I'm actually rather happy that this logic is flattened out, now.

tomcombriat:
So far only had time to look briefly, looks quite neat but haven't done any testing yet.

Point of organisation: if I produce some config files for some other configurations, should I had them to this? If yes, how do you prefer to do? Me merging straight into `devel/Mozzi2` or requesting a PR from another branch?

tfry-git:
> So far only had time to look briefly, looks quite neat but haven't done any testing yet.

Take your time, I won't be all that fast, the upcoming days, either. I also intend to check at least some of those boards - and take a more serious look at your fixmath, of course.

> Point of organisation: if I produce some config files for some other configurations, should I had them to this? If yes, how do you prefer to do? Me merging straight into devel/Mozzi2 or requesting a PR from another branch?

I don't think we need to overdo it with PRs. Commit right away. However, until this is merged, it's in devel/config_rework. devel/Mozzi2 is just the target branch.

internal/config_checks_rp2040.h
```diff
@@ -0,0 +1,113 @@
...
+ *   - Wavetables and samples are not kept in progmem on this platform. While apparently speed (of the external flash) is not much of an issue, the data always seems to be copied into RAM, anyway.
+ *   - Note that getAudioInput() and mozziAnalogRead() return values in the RP2040's full ADC resolution of 0-4095 rather than AVR's 0-1023.
+ *   - twi_nonblock is not ported
+ *   - Code uses only one CPU core
+ *
+ * @section rp2040_output Output modes
+ * The following audio modes (see @ref MOZZI_AUDIO_MODE) are currently supported on this hardware:
+ *   - MOZZI_OUTPUT_EXTERNAL_TIMED
+ *   - MOZZI_OUTPUT_EXTERNAL_CUSTOM
+ *   - MOZZI_OUTPUT_PDM
```
tomcombriat, line 23: Is there really a PDM mode for the RP2040? Did not remember it and could not find it…

tomcombriat:
The amount of work there is impressive!

Starting testing on the RP2040 (it was on the breadboard)… Worked straight away in standard config, minor correction to get the I2S to work.

Just a question on documentation: to find the default settings, and find their names, I had to go to `internal/config_checks_rp2040.h` which I found quite fast especially as I knew what I was looking for. But I wonder if that's a good place to store these settings: `internal` sounds like something standard users will not look into at first glance, and `checks` sounds like checks, not config. I wonder if splitting these files into one `check` actually checking the config is correct and including a `config_default_rp2040.h` (or another name as not all the parameters listed are used as default) which would be more accessible. This file would actually be "the" and a good reference for all the modes and config available for a given platform? Split into section with, for instance the DMA settings said plainly (with a warning on changing them).
Or maybe the plan would be to make a config file for every possible configuration?

tfry-git:
"internal" may not be the best name, but as a side-quest, I'd like to separate headers that are meant to be used in Mozzi, internally, from those that would actually be included in a sketch. And even if these are config-related, they are not a config to include in a sketch (contrary to the config examples), so that's how I ended up with "internal".

I also sort of hope that standard users will not have a need to actually read these files, because the defaults are - and hopefully remain - documented (probably we should place links to that documentation quite liberally). Actually, I must admit, they may not be as pleasant to read, as I had hoped, importantly, because I could not find a workable one line (macro) way of saying: "if XY is not defined, then define it as Z".

As for separating the "checks" from the "defaults", I found those somewhat interrelated, such as "if mode is PDM via I2S, set default pinout to xy, _and_ enforce that only mono is supported". It is entirely possible to split that, but I'm not immediately convinced that this would help readability, overall. No objections to simply renaming config_checks_x.h to config_default_x.h, though. It may also be possible to structure these files better.

> Or maybe the plan would be to make a config file for every possible configuration?

Yes, that's actually the (mid-term) plan. Ok, perhaps at least one config example per platform, and possibly as much as one per audio mode per platform. Part of the idea would be that the config examples should be ones that are actually known to work. Sort of as an intermediate level between "stick with the defaults", and "here's a list of all the things you can customize".

tomcombriat:
> Ok, perhaps at least one config example per platform, and possibly as much as one per audio mode per platform. Part of the idea would be that the config examples should be ones that are actually known to work. 

Sounds good. I think it will very depend on how the documentation is done in the end and I will try to take my share in that. 

(I was going to ask about porting made easy for other contributors but just saw the template actually…)

tomcombriat:
Noting here something I just noticed, and which could (or not) be a bug but certainly worth documenting somewhere.

 If ones wants `SPI` to work one the same core than Mozzi, then `SPI.begin();` and ` SPI.beginTransaction();` should be placed *after* `startMozzi()`. Did not investigate much why…

config/config_example_avr_legacy_standardplus_mode.h
```diff
@@ -0,0 +1,18 @@
+/* Configuration example
+
+This example is targetted at the AVR platform (Arduino Uno & friends), only!
+
+The configuration formerly known as STANDARD_PLUS is still the default on AVR, so you
+do not need to configure anything! This file just lists the relevant settings involved: */
+
+#include "MozziConfigValues.h"  // for named option values
```
tomcombriat, line 8: If I understand well, this is actually not needed in any example right?

tomcombriat:
(Have to say I am quite amazed that you manage to pull such a thing…)

I am leaving comments in the code as I go, hope it is visible. And commits when I see things while testing. I might try to through it all after all the testing I can do is done but I like starting with the experimental approach!

config/mozzi_config_documentation.h
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * Supported values:
+ *   - MOZZI_OUTPUT_PWM Output using pulse width modulation (PWM) on a GPIO pin. This is the default on most platforms.
+ *                      On the Arduino Uno (more generally ATMEGA328P), this allows for a sample resolution of 488 (almost 9 bits) on pin 9.
+ *                      Usable pins and resolution will be different on other boards.
+ *   - MOZZI_OUTPUT_2PIN_PWM Output using pulse width modulation on two GPIO pins, where one pin represents the lower bits, and the other the higer bits of the sample.
+ *                      On the Aduino Uno, this allows for 14 bits of resolution on pins 9 (low) and 10 (high). For further information (wiring etc.) see @ref hardware_avr_2pin.
+ *   - MOZZI_OUTPUT_EXTERNAL_TIMED Output is not controlled by Mozzi itself, but left to the user sketch. This setting allows to completely customize the audio output, e.g.
+ *                      for connecting to external DACs. For more detail, see @ref external_audio
+ *   - MOZZI_OUTPUT_EXTERNAL_CUSTOM As above, but additionally bypassing Mozzi's sample buffer. For more detail, see @ref external_audio
```
tomcombriat, line 57: I understand it makes sense to have an `EXTERNAL_CUSTOM` with respect to the fact that we sometimes bypass the buffer for some output options (directly in Mozzi) but, out of curiosity, does such a thing as ever been employed in an external fashion?

MozziConfigValues.h
```diff
@@ -0,0 +1,44 @@
+/** This file keeps a list of named configuration values.
```
tomcombriat, line 1: Have to say that, for the time being, I do not really get why this file is useful, I might be a bit slow… Is it anticipating for the single compilation unit?

config/mozzi_config_documentation.h
tfry-git: Maybe not thus far, but it seems to make sense in that:
- it's totally easy to implement ;-)
- when interfacing with third-party output libs / mechanisms, it seems quite likely that some from of buffering / rate control is already in place, there.

MozziConfigValues.h
tfry-git: In a way, yes: With a single compilation unit, it will be possible (but not necessary) to specify config options directly at the top of a sketch. Since that needs to happen *before* including MozziGuts.h, we need some place to put the "labels" for the options.

config/config_example_avr_legacy_standardplus_mode.h
tfry-git: Strictly it is not needed, but again - after single compilation unit - a user may want to put ```#include <some_config_example.h>``` near the top of their sketch (and before ```#include <MozziGuts.h>```). Then the config value labels need to be known at that point, and thus MozziConfigValues.h is included.

internal/config_checks_rp2040.h
tfry-git: Mixed it up with PWM. Fixed. (Although I always wanted to add a PDM mode for the RP2040, one of those days...)

tfry-git:
Ok, so I finally tested with STM32 (both cores), ESP8266, and ESP32 boards. No fixes were needed, so I take that as an encouraging sign. I never owned an SAMD-based board, so I guess we'll have to skip testing that on actual hardware.

As things tend to go, these last few bits have taken far too long, and I'm afraid, I may have lost track a little, but I believe all known problems and concerns ought to be fixed/addressed, now? If not, please remind me. Otherwise I'd like to merge this (into the Mozzi2-branch) soon.

tomcombriat:
I think it looks good!

> I finally tested with STM32 (both cores),

I planned to do that but got completely dragged into something else in the meantime (and I never know which one of my stm32 are actually working… Sourcing them was quite hard at some point, do not know how it is now). Maybe allow me a couple more days to test the external on STM32 (I think I tried external on AVR, do not know why it is not in the table), and then this is good to go I think!

tomcombriat:
Quickly reporting: STM32 using the Maple (Clark) core does not seem to work in external mode (too bad, that's what I am using on some legacy hardware…). 

The other core seems to be okay (to be completely confirmed).

tomcombriat:
All good for me!

tfry-git:
Wonderful, I'll merge into devel/Mozzi2 then, before you get a chance to change your mind ;-)

tomcombriat:
I just wanted to test as much as possible with the hardware I have available, sometimes I just need a small push ;).

tfry-git:
> I just wanted to test as much as possible with the hardware I have available, sometimes I just need a small push ;).

That wasn't meant to criticize, it rather felt like I had been the one to drop the ball, and now I was eager to finally merge this. (I also tend to need a small push, BTW)

tomcombriat:
> That wasn't meant to criticize

I did not take like that at all ;)

tfry-git:
It's generally safe to delete branches after merging (full history is still available, anyway). I'm just no good at tidiness...

Issue #207, opened 2023-10-25
Improving fixMath


tomcombriat:
So, the fix maths also came as one point Mozzi could be improved on. Of course, as the majority of the boards Mozzi is supporting do not have an FPU, using fix maths is the way to go to keep efficiency. But fix maths can be a bit disturbing for new users, all these new types, that you need to take care of when doing operations and which display incorrectly on `Serial.print`…

Here are a few suggestions we might think of. This is also a place for discussion.

- documentation: there are some documentations about the numerous fix point types supported by Mozzi, but I think a more fundamental documentation about what are fix point types, how to handle to handle them and so on, might be helpful. Some questions remaining are: where to put that? On the website for instance in another topic like the tricks to make the code faster for instance?
- adding a "smart" multiplication would also be helpful I think. For that I was thinking of implementing a (quite huge) set of functions on the basis of: `fixMult<T1>(T2,T3)` where `T1` being the output type, already shifted the correct way. Here there are a few questions: 
  - these might be implemented by hand (or with a script of some sort, I am not afraid for that) but might not be handled by the compiler as a regular template, so why use one and not `T1 fixMult(T2, T3)`? My personal opinion for the template is to force the user to specify the output type, and not rely on the compiler to decide like it would be the case with a non-templated function without an explicit cast from the user but I am happy to have external opinions.
  - even though it can be written automatically, some cases are not completely straightforward: should we implement for instance the multiplication of an unsigned fixmath with a signed one? 
  - also, I wonder if someone would have a solution to make the compiler do this job via templates or macros?
  - finally, is it a good idea?

Cheers,

tfry-git:
I believe it should be possible to even go one step further and replace (or supplement) the QXnY typedefs with proper classes. For the purpose of this discussion, let's call the replacement type ```NXnY``` (which I don't think will be the final name).

Internally these would simply carry an integer (8, 16, or 32 bits), but importantly they would have operator overloads so, you can simply write: ```
N16n16 number a = ...;
N8n8 number_b = ...;
number_a += number_b;
```. The compiler keeps track of what type each is, and do the appropriate thing, without ever have to spell out the types a second time.

This would also effectively do away with the complexity of specialized members in other classes such as ```Oscil::setFreq_Q16n16()``` etc. That would instead become an overload ```Oscil::setFreq(N16n16 freq)```.

[...]

PR #212, opened 2023-11-15
FixMath2.0

tomcombriat:
This is my results of trying to get a nice FixMath working. This is still working in progress: it is quite messy, ~only `unsigned` is done and except for casts, only the multiplication is implemented~ but that gives the idea and allow for some discussion before I go further.

(@tfry-git you asked for templates, well you'll be served, and macros too ;) ).

---

## In brief:

This defines a new fixmath type2 for unsigned: `UFixMath2<N,n>` and `SFixMath2<N,n>` where `N` is the number of integer bits and `n` the number of fractional bits. 

These two values can be anything, but value safety is only ensured up to 63 bits total (I think this is enough). A static assert prevents going further.

Any `UFixMath2<N,n>` can be casted into any `UFixMath2<N2,n2>`, but if the initial value cannot be represented in the new type you are running into overflow, just like with standard types.

After fiddling for quite some time, I changed the paradigm from what was discussed in #207: the multiplication between  `UFixMath2<N,n>`  and  `UFixMath2<N1,n1>` returns a  `UFixMath2<N+N1,n+n1>`. Basically, (after nearly skipping a night on it), I came to the conclusion that if you return to the first type of the multiplication (as discussed in #207) you have to either:
- shift before, which does not always turn out well if you have types of very different bytes depth (I can expand on it if you want)
- shift after, but in that case you already promoted the integers which are used for internal calculation, so it is a bit sad to remove this precision as the cost as already been paid.

The current paradigm allows for optimisation if you choose the input types so that it fits, but with no margin. The results are as accurate as they can be and won't overflow. If you do not need precision you can always cast them lower prior to multiplication (faster), or after if you want something precise but with fewer bits: 

```
UFixMath2<16,16> toto(200.25);
UFixMath2<8,8> tata(10.5);

UFixMath2<24,24> res1 = toto*tata; // returns 3103.87 printed as float
// eq
auto res1 = toto*tata; 


UFixMath2<16,0> res2 = toto*tata; // returns 3103.00 printed as float, all the fractional bits have been removed
```
In the example above, the choice of the type for `toto` is not smart at all as the value would fit in 8 bits integral part which is why the return type is sub-optimal.

So these types are quite easy to optimize if you know the expected range of your values, then you can choose the smallest container that accommodate them, with the number of fractional bits setting the precision.

### Methods (so far)
- constructors from `float`, `double`
- constructors from another `UFixMath2`
- constructors from an integer
  - interpreted as a plain integer `UFixMath2<8,8> tata(10);`
  - interpreted as a fix typed value in the correct representation: `UFixMath2<8,8> tata(10,as_frac=true);`
- `UFixMath2.asFloat()` returns the value as float
- `UFixMath2.getInt()` returns the internal integer.

The internal integer is using `IntegerType.h` to find the correct container. The formula is a bit weird: `typedef typename IntegerType<((NI+NF-1)>>3)+1>::unsigned_type internal_type ;`, but the `-1, +1` ensures that your container is big enough.

Other methods will to be implemented if we agree with the current layout already. On that, can we assume the compiler to be smart enough to transform `int a/2` into `int a>>1` and skip the shifting operators? I read somewhere that it should but if someone knows for sure. Integration into Mozzi, especially for the `Oscil.setFreq()` are also on the list.

I am also wondering if the division between the FixPoints should be implemented. Once again, assumptions need to be made to shift left the left hand side of the division. Or provide a crude division that does not do it (and let the users cast up if needed). I do not think I ever used fixed point division in any of my Mozzi codes though, might mislead more than help. Of course, division by an integer is absolutely needed but the case is nearly trivial if we assume no shift before.

It is a bit messy for now, tried quite a few things last night on the pre-shift way before convincing myself that this is not satisfying. Will clean up soon. 

Of course, some standard types (like Q8n8) will be typedef'd in the end.

Welcoming comments and suggestions as always!

---

## Planned and acheived

- [x] Signed FixType (might be possible by macro, but not sure it is worth the cipher that will come out of that…)
- [x] multiplication/~division~ with standard integers for any FixType, **not safe**
- [x] multiplication between any FixType, **safe**
- [ ] ~divisions with FixType~ not sure it is actually wise, not planned
- [x] Addition, subtraction between any FixType, **safe**
- [x] Addition, subtraction between anythings , **not safe**
- [x] Right and left shift ** not safe**
- [x] Safe shifts with templated functions
- [x] Check for not asking for too big types 
- [x] Interaction with Mozzi (setFreq in particular)
- [x] Documentation
- [x] Comparisons between FixMath types.
- [x] Inverse, fast (conservative) and accurate of a FixMath
- [x] General Inverse (arbitrary precision)
- [x] Auto constructors from integers to pure fractional or pure integer fixed (to use with caution with "non-typed" parameters)  
- [ ] Adapt Line class (straightforward if we add the division but…)
- [ ] Adapt examples

## Notes

Any operations between FixType is supposed to be safe, in terms of overflow and precision. However, any operation with other types, like `int` *is not*. This is basically to avoid these types to get promoted to easily for instance:
```
UFixMath2<8,8> tata = 10.2;
auto toto = tata*2; // would lead to toto being UFixMath2<24,8> as 2 will be considered `int`
```
I think this will push to use only FixType which, might not be a bad thing as long as people do not abuse on the container size because they can't overflow.

One idea for that might be to allow automatic construction from `const int` (and affiliated), where the constructor (or an affiliated function) determines the `NI` automatically to the smallest needed value. Ideally, it would be neat to do:

```
UFixMath<> tata = 2;
```
and the compiler finds itself that NF=2, but not sure it is possible.

--- 

## Example code (without interaction with Mozzi yet)

```
#include <FixMath2.h>


SFixMath2<8, 8> sf1 =-127.5;
SFixMath2<8, 8> sf2 = 120;
UFixMath2<8,8> uf1 = 16.1;
UFixMath2<16,0> uf2 = 10000;
SFixMath2<8,8> sfconv = uf1;
UFixMath2<8,8> ufconv = sf2;

void setup() {
  Serial.begin(9600);
}

void loop() {
Serial.print(uf1.asFloat());
Serial.print("(UFixMath2<8,8>) *");
Serial.print(uf2.asFloat());
Serial.print("(UFixMath2<16,0>) =");
Serial.println((uf1*uf2).asFloat());

Serial.print(sf1.asFloat());
Serial.print("(SFixMath2<8,8>) *");
Serial.print(uf1.asFloat());
Serial.print("(UFixMath2<8,8>) =");
Serial.println((sf1*uf1).asFloat());

Serial.print("SFixMath(UFixMath(16.1))=");
Serial.println(sfconv.asFloat());

Serial.print("UFixMath(SFixMath(120)) =");
Serial.println(ufconv.asFloat());

Serial.print(sf1.asFloat());
Serial.print("(SFixMath2<8,8>) +");
Serial.print(uf1.asFloat());
Serial.print("(UFixMath2<8,8>) =");
Serial.println((sf1+uf1).asFloat());

Serial.print(sf1.asFloat());
Serial.print("(SFixMath2<8,8>) -");
Serial.print(uf1.asFloat());
Serial.print("(UFixMath2<8,8>) =");
Serial.println((sf1-uf1).asFloat());

Serial.print(sf2.asFloat());
Serial.print("(SFixMath2<8,8>) -2 =");
Serial.println((sf2-2).asFloat());

Serial.print(sf1.asFloat());
Serial.print("(SFixMath2<8,8>) -1 =");
Serial.print((sf1-1).asFloat());
Serial.println(" (Overflow)");

Serial.print("2-");
Serial.print(sf2.asFloat());
Serial.print("(SFixMath2<8,8>) =");
Serial.println((2-sf2).asFloat());

Serial.print("10-");
Serial.print(uf1.asFloat());
Serial.print("(UFixMath2<8,8>) =");
Serial.print((10-uf1).asFloat());
Serial.println(" (Promotion to SFixMath2<9,8>");

UFixMath2<8,8> tt = 250;
auto tt2 = -tt;
Serial.print("- (UFixMath2<8,8> tt = 250) = ");
Serial.print((tt2).asFloat());
Serial.println(" (Promotion to SFixMath2<9,8> with auto)");

Serial.print(uf1.asFloat());
Serial.print("(UFixMath2<8,8>) >>2 = ");
Serial.println(((uf1>>2)).asFloat());

Serial.print(uf1.asFloat());
Serial.print("(UFixMath2<8,8>) <<4 = ");
Serial.print(((uf1<<4)).asFloat());
Serial.println(" (Overflow)");

Serial.print(uf1.asFloat());
Serial.print("(UFixMath2<8,8>).sR(2)= ");
Serial.print((uf1.sR<2>()).asFloat());
Serial.println(" (Demotion to UFixMath2<6,8>)");

Serial.print(sf1.asFloat());
Serial.print("(SFixMath2<8,8>).sL(2)= ");
Serial.print((sf1.sL<2>()).asFloat());
Serial.println(" (Promotion to UFixMath2<10,8>)");

Serial.println();

Serial.println();


  delay(1000);
}

```


tfry-git:
Looks promising! I will need time to look at it in more depth, so just a first set of thoughts:

- No great idea, so far, but I think we should think about how to avoid confusion between "int" (the regular number), and "int" the underlying storage type in UFixMath. I think ```getInt()``` is not an ideal choice, here, perhaps one ```asInt();``` (same as ```asFloat()```, and one ```asRaw()```? Again, not sure on the name, but I think asRaw() would prevent somebody using it under a wrong assumption of "oh, I know what 'int' does."
- Constructor: In a similar vein, I think that may be too error-prone, or too hard to read the code late. I suggest splitting out the "raw" constructor, e.g. into a static "factory" function (inside the class, so it can initialize the private storage: ```static UFixMath2<NI,NF> fromRaw(type raw);```
- Multiplication: I agree with your reasoning. However, should there be a limit to the upshifting (when the are many multiplications)? Won't down-cast before multiplication cause extra instructions? And won't it discard precision prematurely, in case the result would actually fit? Note that - as an idea - it should not be too hard to add a customizable mutliplication as a non-operator member function: ```template<byte NIout, byte NFout, byte NIother, byte NFother> UFixMath<NIout, NFout> mult(const UFixMath<NIother, NFother> other);``` (declaration looks terrible, but function calls would be as easy as ```toto.mult<16, 16>(tata);```
- Division: No real idea, yet.

Regarding what code gets generated: Do you know godbolt.org? It's great for this sort of task, and also has AVR gcc (be sure to specify the "-O2" compiler option, manually!).

tomcombriat:
Thanks for your comments!

> No great idea, so far, but I think we should think about how to avoid confusion between "int" (the regular number), and "int" the underlying storage type in UFixMath. 

Good point. I will change that. Note that in my mind, to only get the integer part, one could have casted to a type with no fractional part but, while this will still be possible, I agree that `asInt()` is misleading as it is now.

> nto a static "factory" function (inside the class, so it can initialize the private storage: static UFixMath2<NI,NF> fromRaw(type raw);

I am not sure I followed you completely here but I think I do, will change also to be coherent with the first comment.

> I agree with your reasoning. However, should there be a limit to the upshifting […]?

Maybe, but note that there are no limits for standard integers, you are free to overflow them at will ;). That being said, it should not be too hard to add some boundaries, for instance with a `static_assert` here, or in `IntegerType` (I tend to think it would make more sense there actually). Will try!

For the down cast, the good thing here is that you never need to, and it won't by default. But in some codes, it is needed to downcast before in order to keep the types down. Maybe a bad example but here is a snip of one of my synth source code:
```
int oscil1 = (((aSin_next * (255 - mix1) + aSquare_next * (mix1)) >> 8 ) * (255 - mix_oscil)) >> 5 ;
```
Downcasting only at the end would lead to an overflow, so it is needed to cast before to cut some bits as we go. But both are possible with the current solution, by default you can keep all precision and be safe (up to the limit of the board), or you can cast down, in fix type which is the neat thing, and control the size of what you end up with. It also allows (old types also, but I think this makes it clearer) to keep the precision only where it matters.

> Division: No real idea, yet.

I have got an idea there, we'll see how that turns out.

Did not have time to work on it today, but will try to get a more complete version this weekend. I'll update my initial post as I go as a mock-up documentation.

 --- 

> Do you know godbolt.org?

No, I did not, but that looks super useful for exploring these kind of optimizations, thanks!

--- 

That might be a bit further down the line, but I am starting to think that, if this turns well, it could also be useful outside of Mozzi. So might be worth thinking of publishing it also as a separate library. Mozzi could embed a copy of it, or call it, depending what is the easiest… Once again, not sure we are there yet…

tomcombriat:
Hi,

I took into consideration your suggestions. I still have a doubt if I understood your "split the constructor" [correctly](https://github.com/sensorium/Mozzi/blob/8681b646d16e19813702d4b6e7b971afa24a6668/FixMath2.h#L44C1-L44C1).

I also added an multiplication overload for "standard" types, like multiplying with an `int`, we multiply the internal value. I choose not to prevent potential overloads in these case to avoid too much promotion (with `int` used as default type, this adds 16bits with any multiplication). If you want to be safe, use only FixMath would be my motto but I feel this is a subjective choice and happy to discuss it.

On that I ran into an unforeseen problem that you might have a neat solution at hand: I had no problem [defining the overload `UFixMath2*T`](https://github.com/sensorium/Mozzi/blob/8681b646d16e19813702d4b6e7b971afa24a6668/FixMath2.h#L81) (with `T` a typename), but the reverse confuses the compiler, even when done outside the class as a free function like:
```
template<typename T,byte NI, byte NF>
     UFixMath2<NI,NF> operator* (const T op, const UFixMath2<NI, NF>& uf)
  {
    return UFixMath2<NI,NF>(uf.asRaw()*op,true);
    }
```

Seems that the compiler sees an ambiguity between the cast from any type `T` to a FixType and this operation, leading to me defining (only a few for now) [specifications for all the types](https://github.com/sensorium/Mozzi/blob/9a4d04c53278b83f939fd77951e50b395acf3a23/FixMath2.h#L129), which is not satisfying. Am I missing something? Thanks!

tfry-git:
This is just a theory, I'm not even sure, I understand the problem correctly (what exactly do I need to paste into the compiler to produce the problem?). Anyway, I think it may be the int constructor that is causing the trouble. Since the second parameter has a default value, this c'tor would allow the compiler to cast int to a UFixMath.

If that is the case, one way would be to declare all constructors "explicit". Another - and this is what I had in mind earlier - is to formally avoid constructors, but rather (not sure, my syntax is 100% correct):

```
template<byte NI, byte NF> // NI and NF being the number of bits for the integral and the fractionnal parts respectively.
class UFixMath2
{
[...]  
private:
  UFixMath2() {}
public:
  static UFixMath2<NI, NF> fromRaw(int raw) {
    UFixMath2<NI, NF> ret;
    ret.internal_value = raw;
    return ret;
  }
  static UFixMath2<NI, NF> fromInt(int integer) {
    UFixMath2<NI, NF> ret;
    ret.internal_value = integer << NF;
    return ret;
  }
  [...]
}
```

And then you'd write:
```
auto tata = UFixMath<8, 8>::fromInt(2);
```

for construction.

Of course a hybrid approach should work as well: "Regular" constructors declared explicit, and the "raw" constructor moved to a static factory function.

tomcombriat:
Hi again,

Sorry I did not give a precise test, indeed I think this is the problem too.

Okay I understand more what you meant earlier, thank!

I feel a bit bad to have to leave out the syntax (which might be ambiguous though): `UFixMath2<8,8> tata(10);` which would have been neat. I think the problem with the mixed approach is that, for multiplying say with standard `int` would ideally work on the raw data…

Thanks for your advices, I'll try to come up with a neat solution building on that! The way I feel now would be to move only one of them outside (probably the `asRaw`) to keep the syntax light for the user, but that won't resolve this ambiguity. Out of curiosity, except that it might be hard to maintain and a bit inelegant, is there anything very wrong about expliciting all the reverse operators (their syntax is [very light](https://github.com/sensorium/Mozzi/blob/9a4d04c53278b83f939fd77951e50b395acf3a23/FixMath2.h#L130C1-L130C1)).

Thanks again!

tomcombriat:
I'm actually starting to think that using [this library](https://github.com/Pharap/FixedPointsArduino) straight might actually be the cleanest instead of recoding a whole new thing, maybe by putting it as required library upon Mozzi installation from the IDE. As of now,  we would have license problems but that might change soon?

I haven't tried it though

tfry-git:
> is there anything very wrong about expliciting all the reverse operators

I'm not sure, what this implies, but if it makes things work, why not. I'm simply more used to "explict" on constructors.

>I'm actually starting to think that using [this library](https://github.com/Pharap/FixedPointsArduino) straight might actually be the cleanest

I cannot currently estimate, how much code (and work) your solution would mean. Avoiding an extra dependency would have some value, too. But in fact, if that means many hours of work, it may not be worth it.

Not that we probably don't have to be as complete as that other library (implementing logic operators and all that jazz). As long as we do have a basic UFixMath-type, it would still be easy enough - for those who really wish to, to interface with a more complete library.

tomcombriat:
Thanks!

I am not really afraid of the work involved, I think it is a great thing for me to get more familiar with this kind of programming. I am trying to come up with something really targeted toward performances for Mozzi, depending how that turns out we can switch to the other one if needed!

tomcombriat:
Alright, I start to like the way it behaves so, before extending the part of Mozzi which could leverage this, maybe it is good to be all on the same page.

I have added some documentation inside the source file, and a test code in the initial comment of this discussion. 
The implementation is sometimes a bit verbish, you might have ways to shorten that, but I did not find how to do a few things more elegantly while keeping the way I wanted to work. The idea is to have something as close as possible to standard types (float actually). So, as said in the documentation, all permitted operations between Fix types are safe and keep precision. The only thing one should take a bit care is not to cast unneededly with too many unnecessary bits. 

At the end of the day, I do not think the user will actually to cast so many of these (`Oscil.setFreq()` but usually called with `Q16n16_mtof` and casting a UFixMath2<16,16> from a MIDI note is straightforward). 
Of course, clean examples will be added in due time.

FixMath2.h
```diff
@@ -0,0 +1,845 @@
...
+ - all operations between fixed point number is safe (it won't overflow) and preserve the precision. In particular:
+   - only addition, subtraction and multiplication are implemented
+   - any operation between a signed and an unsigned leads to a signed number
+   - resulting numbers will be casted to a type big enough to store the expected values. It follows that it is worth starting with types that are as small as possible to hold the initial value.
+ - all operations between a fixed point number and a native type (int, float, uint) are *not* safe. If the resulting value cannot be represented in the fixed point type it will overflow. Only addition, subtraction, multiplication and right/left shift are implemented.
+ - safe right/left shifts, which return the correct value in the correct type are implemented as .sR<shift>() and .sL<shift>() respectively, shift being the shifting amount.
+*/
```
tfry-git, line 30: This should eventually contain some code examples.

If I understand correct, the advice will be to use as little a number for NI as possible, and not necessarily 8-bit aligned. For NF, I believe an 8bit-aligned number will be preferable, as this a) this does not grow, and b) shifts by 8 are more efficient on AVR. (If this is correct, it should also be stated).

FixMath2.h
```diff
@@ -0,0 +1,845 @@
...
+template<byte NI, byte NF>
+class SFixMath2;
+
+
+
+/** Instanciate an unsigned fixed point math number.
+@param NI The number of bits encoding the integer part
+@param NF The number of bits encoding the fractional part
+*/
+template<byte NI, byte NF=0> // NI and NF being the number of bits for the integral and the fractionnal parts respectively.
```
tfry-git, line 77: Default to 8 for NF? 0 seems a bit pointless.

FixMath2.h
```diff
@@ -0,0 +1,845 @@
...
+    if (as_raw) fromRaw(value);
+    else internal_value = (internal_type(value) << NF);
+  }
+
+
+  /** Set the internal value of the fixed point math number.
+@param raw The new internal value.
+  */  
+  template<typename T>
+  void fromRaw(T raw) { internal_value = raw; }
```
tfry-git, line 121: Can this be made static? That would allow easier use:

```
auto a = UFixMath<x,y>::fromRaw(z);
// rather than
UFixMath<x,y> a;
a.fromRaw(z);
```

FixMath2.h
```diff
@@ -0,0 +1,845 @@
...
+@param uf An unsigned fixed type number which value can be represented in this type.
+@return A unsigned fixed type number
+  */
+  template<byte _NI, byte _NF>
+  UFixMath2(const UFixMath2<_NI,_NF>& uf) {
+    internal_value = SHIFTR((typename IntegerType<((MAX(NI+NF-1,_NI+_NF-1))>>3)+1>::unsigned_type) uf.asRaw(),(_NF-NF));
+  }
+
+  /** Constructor from another SFixMath2. 
+@param uf An signed fixed type number which value can be represented in this type.
```
tfry-git, line 137: So sign is discarded, right? State that.

FixMath2.h
```diff
@@ -0,0 +1,845 @@
...
+  //////// ADDITION OVERLOADS
+
+  /** Addition with another UFixMath2. Safe.
+@param op The UFixMath2 to be added.
+@return The result of the addition as a UFixMath2.
+  */
+  template<byte _NI, byte _NF>
+  UFixMath2<MAX(NI,_NI)+1,MAX(NF,_NF)> operator+ (const UFixMath2<_NI,_NF>& op) const
+  {
+    constexpr byte new_NI = MAX(NI, _NI) + 1;
```
tfry-git, line 155: I was a bit hesitant about the +1 at first. int addition does not do that, after all. But I guess it makes sense, esp. in combination with the earlier comment that NI does not need to be divisible by 8.

FixMath2.h
```diff
@@ -0,0 +1,845 @@
...
+  }
+
+
+  //////// SUBSTRACTION OVERLOADS
+
+       /** Subtraction with another SFixMath2. Safe.
+@param op The SFixMath2 to be subtracted.
+@return The result of the subtraction as a SFixMath2.
+  */ 
+  template<byte _NI, byte _NF> // We do not have the +1 after MAX(NI, _NI) because the substraction between two SFix should fit in the bigger of the two.
```
tfry-git, line 521: Not necessarily true in the signed case.

FixMath2.h
```diff
@@ -0,0 +1,845 @@
...
+  */
+  template<typename T>
+  UFixMath2<NI,NF> operator* (const T op) const
+  {
+    return UFixMath2<NI,NF>(internal_value*op,true);
+  }
+
+
+  /** Right shift. This won't overflow but will not leverage on the bits freed by the shift.
+Better to use .sR<shift>() if possible instead.
```
tfry-git, line 240: If I understand correctly, you implemented .sR() separately, after all. Why not as a member?

tfry-git (commented):
Ok, here's a first round of comments. I did not dive in super-deep so far, and importantly, I did not actually test anything (I suppose it should even be possible to formalize much of the testing as static_asserts?).

Looks pretty sold on first impression, though.

Probably ok to rename as U/SFixMath (without the "2")?

FixMath2.h
```diff
@@ -0,0 +1,845 @@
+/*
+ * FixMath2.h
+ *
+ * Copyright 2023, Thomas Combriat and the Mozzi consortsium
```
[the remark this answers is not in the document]
tomcombriat: Will do.

FixMath2.h
```diff
@@ -0,0 +1,845 @@
...
+
+
+
+
+#ifndef FIXMATH2_H_
+#define FIXMATH2_H_
+
+#include "IntegerType.h"
+
+#define SHIFTR(x,bits) (bits > 0 ? (x >> (bits)) : (x << (-bits))) // shift right for positive shift numbers, and left for negative ones.
```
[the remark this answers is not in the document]
tomcombriat: I think I am for making them more longish, `SHIFTR` could actually be useful in other places maybe.

FixMath2.h
tomcombriat: Actually, both should be as small as possible. As the integral part and the fractional part are contained inside one integer under the hood, I do not think 8bits aligned numbers are preferable for the fractional part but of course it is preferable to have the underlying int as small as possible and aligned (NI+NF = 9 forces to go 16bits under the hood). Of course if 8 shifts are more efficient on AVR that could be beneficial but such shifts occurs only at construction from a "pure" integer or float, which should not happen very often. On the long run (but I need to implant this into the relevant part of Mozzi) a few pre-defined types will be given. Note that `NF` grows with multiplication.

FixMath2.h
tomcombriat: That's a remain of trying to make an automatic constructor for integers which would find the smallest working `NI` and putting `NF=0`. Unfortunately I did not manage to make that work (too bad macros cannot be recursive). So I think I'll remove this default value as a whole, it is more asking for wrong usage than anything else.

FixMath2.h
tomcombriat: Oh yes, good catch. I have to say that `static` is not my most familiar C++ concept so I usually do not think of it. Will do.

FixMath2.h
tomcombriat: Yes, basically this enforces that no overflow can happen, like the behavior of a float (well, to some extend). It really behaves differently from standard integer types but I think that, wisely used and well introduced it can really be neat to not care too much about overflows once things are created.

FixMath2.h
tomcombriat: Wooooooooooops, Very good catch!!

FixMath2.h
tomcombriat: `sL()` and `sR()` are implemented as members a bit further (l 263).

FixMath2.h
tomcombriat: Following this, I think all the substraction between UFix and SFix should get a `+2` and not a `+1`? If I get my maths right…

tomcombriat: Or even more precisely: `SFixMath2<MAX(NI,_NI+1)+1,MAX(NF,_NF)> operator- (const SFixMath2<NI,NF>& op1, const UFixMath2<_NI,_NF>& op2 )`

FixMath2.h
tomcombriat: Note that I realize now that .SR could be "just" moving the comma and not removing some bits on the right so that (pseudo-code):
```
UFixMath<8,8>.sR() -> UFixMath<6,10>
```
instead of 
```
UFixMath<8,8>.sR() -> UFixMath<6,8>
```
as it is now.

When I implemented it the first behavior seemed logical, but now I have the feeling that it does not align with the otherwise coherent paradigm of ensuring safety and keeping precision. Any opinion @tfry-git ?

Note that, if it is changed, the old behavior can easily be recover by enforcing the output type (conversion is the big plus of these classes, you can reduce precision easily with them):
```
UFixMath<8,8> a;
UFixMath<6,8> = a.sR(); // UFixMath<8,8>
```
instead of using `auto`:
```
auto b = a.sR(); // UFixMath<6,10>
```

I have the feeling a very comprehensive documentation will be useful for avoiding weird usages that goes outside the intended goal. Good thing that now the Arduino IDE does infer the type so the actual type represented by  `auto` is actually inferred in the IDE.

tfry-git: Yes, I think aiming for coherency is indeed the way to go, so it really makes sense to extend it to sR()/sL().

But also, yes, good documentation and example(s) will be key.

A half-baked idea for brainstorming: Suppose we add a third template parameter "max bits", defaulting to 32 or 64. This will be carried along any operation (probably the smaller value, in case of ambiguity). Each operator will additionally get a static_assert to ensure that the total number of bits is still below this, with a hint to add a downcast, somewhere.

FixMath2.h
tfry-git: Alternative way of thinking about it: Don't count the sign bit in NI, but only take it into account when determining the required storage type. I.e. a "plain" int16_t would correspond to SFixMath2<15,0>.

I think that should make the math easier?

FixMath2.h
tfry-git: (Static is also renowned as one of the more confusing concepts of C++, esp. since the keyword has three totally different meanings, depending on where it is used.)

FixMath2.h
tomcombriat: That sounds reasonnable, I'll have to think about it. Note that there are already static_assert for preventing casting something bigger than 64. This automatically kick in if you try to do an operation which result exceed 64bits.

tomcombriat:
Just throwing a small idea here:

Is there anything wrong into moving all of `mozzi_midi.cpp` inside `mozzi_midi.h`? Would that pose some problems with respect to the `midiToFreq` table? Moving it there would allow to be a bit more coherent as I had to move the templated version of `mtof` for `UFixMath` there. It would also allow some more inlining I guess. 

Anyone against?

tomcombriat:
Alright, I think the important things are there. 

This can now be used with mozzi_midi and with `Oscil`. For these two, it falls back to the legacy mozzi_fixath to avoid code duplication. So far untested, will try to do that and to provide nice examples soon.

Best,

tfry-git:
> Just throwing a small idea here:
> 
> Is there anything wrong into moving all of `mozzi_midi.cpp` inside `mozzi_midi.h`? Would that pose some problems with respect to the `midiToFreq` table? Moving it there would allow to be a bit more coherent as I had to move the templated version of `mtof` for `UFixMath` there. It would also allow some more inlining I guess.

Totally agree with this. That's also one step further towards the "single compilation unit". I also think (as in talking without testing), that it should not be a problem for the midiToFreq table. The trouble to watch out for, is what will happen if two separate .cpp files both include a call to mtof(). Will the linker be smart enough to discard one copy of the table? I hope so, but am not 100% sure.

> Edit:
> 
> In the same vein, I guess `phase_increment_fractional = ((unsigned long)frequency) * ((256UL*NUM_TABLE_CELLS)/UPDATE_RATE);` and the likes could be replaced by: `phase_increment_fractional = ((uint32_t)frequency) * ((256UL*NUM_TABLE_CELLS)/UPDATE_RATE);`, which would have no effect on AVR but provide a small boost for 32bitters by avoiding to go to 64 bits?

Good catch!

tfry-git:
> > Just throwing a small idea here:
> > Is there anything wrong into moving all of `mozzi_midi.cpp` inside `mozzi_midi.h`? Would that pose some problems with respect to the `midiToFreq` table? Moving it there would allow to be a bit more coherent as I had to move the templated version of `mtof` for `UFixMath` there. It would also allow some more inlining I guess.
> 
> Totally agree with this. That's also one step further towards the "single compilation unit". I also think (as in talking without testing), that it should not be a problem for the midiToFreq table. The trouble to watch out for, is what will happen if two separate .cpp files both include a call to mtof(). Will the linker be smart enough to discard one copy of the table? I hope so, but am not 100% sure.

Afterthought: Maybe it's worth some extra effort to make the lookup-table private, though, so we remain free to re-design, should the need arise. Suggested solution:

```
class MozziMidiPrivate {
private:
  friend function int mtof(uint8_t); // add other functions using this, here
  static CONSTTABLE_STORAGE(uint32_t) midiToFreq[128];
};

static CONSTTABLE_STORAGE(uint32_t) MozziMidiPrivate::midiToFreq[128] = [...];
```

That should give the linker a definite hint not to duplicate the table, too.

tomcombriat:
Thanks!

Looking more into it, I wonder if the [line](https://github.com/sensorium/Mozzi/blob/a9fd035292cbb428c263a4d76d4fd3562e2f1742/Oscil.h#L230C4-L230C92) 

```
phase_increment_fractional = ((unsigned long)frequency) / (UPDATE_RATE/NUM_TABLE_CELLS);
```

could not benefit from calculating the inverse (`NUM_TABLE_CELLS/UPDATE_RATE` with a preliminary shift on the upper part of the fraction, which would be removed after, for performances purposes. I need to try if that actually makes things faster of if the compiler is smart enough. Then it would ressemble a bit more like the Q24n8 version (with an additional right shift).

tomcombriat:
Just added comparisons between FixMath types. Did not added with other types as they can be converted to the user very easily. 

Actually, I wonder if all the overloads with standards types are actually needed… Users can convert anything to FixMath using the constructors. The drawbacks would be a bit of longer syntax for the user (and might lead to type explosion if they are not set quite small to start with), but at the same time this ensures type safety and overflows. It would also ease quite a bit the implementation file… 

Any opinion?

tfry-git:
> Just added comparisons between FixMath types. Did not added with other types as they can be converted to the user very easily.
> 
> Actually, I wonder if all the overloads with standards types are actually needed… Users can convert anything to FixMath using the constructors. The drawbacks would be a bit of longer syntax for the user (and might lead to type explosion if they are not set quite small to start with), but at the same time this ensures type safety and overflows. It would also ease quite a bit the implementation file…
> 
> Any opinion?

Since these overloads are unsafe, it may in fact be a good idea to remove them. Importantly, I imagine users would often not be aware that something "unsafe" is going on behind the scenes, at all.

I have not really played with fixed math, myself, and I'm not sure, how much of a hassle this will actually mean in the real world. You definitely have more experience, here.

But, on very general grounds, adding convenient features, later, is always possible, while removing features means breaking existing sketches. So I'd recommend to err towards a more reduced implementation, initially.

tomcombriat:
> But, on very general grounds, adding convenient features, later, is always possible, while removing features means breaking existing sketches. 

Good point! I'll convert the examples and see how it turns. I see advantages of both paradigms (completely safe and type preservation), so will see how this turns out.

In all cases, I think good documentation and especially tutorials might be helpful (and not onlyy for this, also for outputting wiring diagrams for instance). If I recall correctly, there are now several places where this kind of things are:
 - "[learn](https://sensorium.github.io/Mozzi/learn/)" in the website
 - "[wiki](https://github.com/sensorium/Mozzi/wiki)" on github
 - "Readme" on github (to some small extend)

I do not know how much the "learn" section is updated automatically alongside the wiki (@sensorium ?) but might be good to agree on one place for additional documentation?

sensorium:
Hi Tom,
 maybe this could be a different thread (about Documentation?), but some quick answers here for now:
    > places where this kind of things are:

        >"learn" in the website
        > how much the "learn" section is updated automatically alongside the wiki 
Ha ha, there's no automatic updating of the learn section, only some very minor updates by extreme force of will through the years.

        >"wiki" on github
I wonder if the (presently empty) wiki could be a way forward...

FixMath.h
```diff
@@ -0,0 +1,1047 @@
...
+
+
+
+#ifndef FIXMATH2_H_
+#define FIXMATH2_H_
+
+#include "IntegerType.h"
+
+#define MOZZI_SHIFTR(x,bits) (bits > 0 ? (x >> (bits)) : (x << (-bits))) // shift right for positive shift numbers, and left for negative ones.
+#define MAX(N1,N2) (N1 > N2 ? N1 : N2)
```
tfry-git, line 69: This may have a high chance of clashing with defines set in other libs. Either rename to MOZZI_MAX, or test ```#ifndef``` MAX before defining. While at it, I also suggest parentheses around N1 and N2 (these are not "passed" as values but rather as mere string tokens, here, so something like ```MAX(a+1, 2)``` will not work as expected.

FixMath.h
```diff
@@ -0,0 +1,1047 @@
...
+#define UBITSTOBYTES(N) (((N-1)>>3)+1)
+#define SBITSTOBYTES(N) (((N)>>3)+1)
```
tfry-git, line 71: I''m a bit uncertain, whether these should also all get a ```MOZZI_``` prefix. I somewhat lean towards that for safety, even if they are certainly much less likely to clash with anything.

tfry-git: (And - quite idly, please don't take this as a request - I'm wondering, whether those could be turned into ```constexpr```s (which would allow placing them in a namespace of something))

FixMath.h
```diff
@@ -0,0 +1,1047 @@
...
+  */
+  template<typename T>
+  UFixMath<NI,NF> operator* (const T op) const
+  {
+    return UFixMath<NI,NF>(internal_value*op,true);
+  }
+
+
+  //////// INVERSE
+  UFixMath<NF,NI> invFast() const
```
tfry-git, line 274: These look really nifty, but I must admit, I don't understand them. Looking forward to your examples!

FixMath.h
tfry-git: D'uh, you're undef'ing them at the end of the file, already. Parentheses in MAX will still be a good idea, though, to avoid future surprises.

FixMath.h
tomcombriat: Will add the parentheses soon, thanks!

FixMath.h
tomcombriat: > - I'm wondering, whether those could be turned into constexprs

Indeed, I thought about it. I remember trying and got deceive by some C++17 requirement but I might have done something wrong. Might try again, it would indeed be more C++ like…

FixMath.h
tomcombriat: Yeah, I thought having the inverse (normally computed in a slow loop, like `updateControl()` could be useful for having fast division, by multiplying by the inverse.

However, like most things I try to do, the great mind thought plan does not turn as expected (and I lose part of my night on it ;)).

Will do a clean description somewhere but here are starters, (which also help me to structure my mind, you don't have to read it all:

The basic idea, instead of dividing one by the number (which would always give zeros with integer) is to do:

((1<<(NI+NF))/a)>>(NI+NF).

Actually, in order to fit in the container, I use 1>>(NI+NF) - 1, but that does not change the matter much. Now for the output sizes:

A FixMath number is ruled by both its *range*, given by NI and its precision, given by NF. These are fundamentally different as increasing NF does not increase the range (which is is [0,1[) by NI does.

The inverse of any number `a`, actually has a reverse relationship: the bigger the initial range, the bigger the resolution of the inverse and vice-versa. This explains the implementation of `invFast()`. With this inverse, both the biggest number, and the smallest one are representable and have correct values: the inverse of the biggest one becomes the resolution and the inverse of the smallest one becomes the range. But there is a small problem: because 1/x is not linear (at all), this leads to a lot of duplicated values: if you take a UFixMath<8,0>, all values between 128 and 255 will have the same inverse, because the resolution is 1/255.

To solve that, one needs to compute the needed resolution so that the inverse of the *biggest number* and the *before biggest* do lead to different results, this is `invAccurate()`. Of course, the size completely explodes doing that so it should be used only when needed (or should be downcasted slightly after). I found the relation of the templates arguments by chance initially but I now have a more or less clean demonstration (actually the relation there is only valid to the first order, but that yields quite good results).

Now I understand why I saw written the division in fixed point arithmetic is tricky, and why floating point is better in that respect. People should have stumbled on the same problems… 40 years ago…

FixMath.h
```diff
@@ -0,0 +1,1047 @@
...
+#ifndef FIXMATH2_H_
+#define FIXMATH2_H_
+
+#include "IntegerType.h"
+
+#define MOZZI_SHIFTR(x,bits) (bits > 0 ? (x >> (bits)) : (x << (-bits))) // shift right for positive shift numbers, and left for negative ones.
+#define MAX(N1,N2) (N1 > N2 ? N1 : N2)
+#define UBITSTOBYTES(N) (((N-1)>>3)+1)
+#define SBITSTOBYTES(N) (((N)>>3)+1)
+#define ONESBITMASK(N) ((1ULL<<(N)) - 1)
```
tomcombriat, line 72: Question for @tfry-git : will this be optimized by the compiler in terms of type (even with the `ULL`?). I mean, `ONESBITMASK(8)` should be: 0b11111111 ideally, but is the `ULL` leading to 0b000000000000000000011111111?
Seems okay on gobolt, but prefer to be sure.

tomcombriat:
Interestingly, it seems that, at least on the Uno, jumping to `float` to compute the inverse and casting back to FixMath is more efficient (2%) than `invAccurate` (but 2.5x slower than `invFast` for the same number of output bit). So might be worth replacing the inside of this `invAccurate` as a wrapper to 
```
1./self.asFloat();
```

Will do more test to confirm this, especially on other platforms. Compilers are getting quite good, code inspection reveals a very similar algorithm than mine when passing via a float…

FixMath.h
tfry-git: > Question for @tfry-git : will this be optimized by the compiler in terms of type (even with the `ULL`?). I mean, `ONESBITMASK(8)` should be: 0b11111111 ideally, but is the `ULL` leading to 0b000000000000000000011111111? Seems okay on gobolt, but prefer to be sure.

That's always difficult to tell, as optimization is always a "may", and never a "must". But I do think it's a fairly safe bet in this case: As long as N is known at compile time, so is the result, and the compiler will surely be smart enough to figure out what data type it fits into. Now, in the case of integer division, that also clearly limits the range of the result of the division. And then, shortly after, the result is cast to internal_type. Of course things would stand quite different, if other operations would be involved.

---

Suggested alterative
```
[inside UFixMath itself:]
private:
  static [constexpr if possible] internal_type onesbitmask() { return ((1ULL << (NI+NF)) - 1); }
```

---

> Another one: should everything in there be inline to avoid clashes with user cpp files in your opinion?

Yes, indeed. More specifically, it's the one-defintion-rule, again, when more than one .cpp file is involved. That should "only" concern the operators, AFAICS, though: Definitions inside a class-declaration are automatically inline.

tomcombriat: > Edit: Or even, avoiding your initial question:
`static [constexpr if possible] internal_type onesbitmask() { return (internal_type(1) << (NI+NF)) - 1); }`

Unfortunately, I think this provokes an overflow if NI+NF=8 or 16 or 32, as the idea is to have all bits set to one, so we take the next value and substract one. A for loop might work in the internal type but I need to see if `constexpr` allows it.

Otherwise, from tests with gobolt, it seems that the compiler do optimize it away.

tomcombriat: > A for loop might work in the internal type but I need to see if constexpr allows it.

It does… After C++14… So no luck there. Your initial proposition do work, but I guess this is more or like the macro in effect.

tomcombriat: `static constexpr` work. Two of them are needed because the accurate division needs a bigger one.

From compiler statistics, it seems to reduce quite a bit the prog mem compared to macros, but I haven't checked which one do perform the best in terms of speed. Committed that for now nevertheless and will try to do some perf check soon. I'll squash a bit my commits at the end, this completely exploded…

tomcombriat: Final word, `constexpr` version is *way* faster… Don't fully understand why but this is striking: 3240us for a thousand `invFast()` original, 2040us for the one using `constexpr`, this is massive!

Will switch it all to `constexpr` and retest fully the computations.

tfry-git (commented):
>  Added constexpr as replacements of macros, but not sure it is better as they will spill out.

You could wrap them into ```namespace MozziPrivate```, (or MozziFixMathPrivate, or place them as private statics into the U/SFixMath classes).

tomcombriat:
> You could wrap them into namespace MozziPrivate, (or MozziFixMathPrivate, or place them as private statics into the U/SFixMath classes).

@tfry-git For the private static, I was a bit afraid that would lead to a lot of duplicates in the generated code by the compiler (hence penalty in flash) as any U/SFixMath needs a lot of different `U/SBITSTOBYTE`, I guess it would be optimized away though, can check. I think the namespace is more the way to go!

tomcombriat:
If you want to have a look, the example I just added was precisely the kind of things I add in mind with FixMath. The main usage is in combination with MIDI and except than the `fromRaw()` that is a bit heavy syntax (but impossible to simplify it if we want to keep the `int` constructor), that might be replaced with `SFixMath<0,7>(aVibrato.next(),true)`, I think it looks fairly clean.

Will try to change some other examples soon.

tomcombriat:
> Some ideas for brainstroming:

And another one:
`(SFixMath<0,7> (osc.next()).sR<7>()`

But not sure it is way better… I was thinking to do a convenient function but that guides toward the having the whole sample in fractional, whereas some situations can very likely benefit from having samples as, say, `SFixMath<5,2>`.

The more logical way, for me would be (my other suggestion in two lines):

```
SFixMath<7,0> toto oscil.next();
auto tata = toto.sR<7>();
```
but will continue thinking about it…

tfry-git:
Another one:

```
template<int NI=0, typename T> auto toSFix(T val) {
    return SFixMath<NI, sizeof(T)*4-NI-1>(); 
}

auto toto = toSFix(aSin.next()); // SFixMath<0, 7>
auto tata = toSFix<5>(aSin.next()); // SFixMath<5, 2>
```

(Alternatively make NF, instead of NI available as parameter. Not sure, which is less confusing.)

tomcombriat:
And yet another (which is inspired from yours but allows for smaller types if needed, for instance `NI=2; NF=4`):

```
template<byte NI, byte NF> auto toSFix(T val) {
    return SFixMath<NI, NF>::fromRaw(val); 
```

I think that's a way to go, especially because the `::` is not encountered very often (and never in Mozzi), so I am a bit afraid of the troubles in can bring to users. Maybe `toSFixRaw` instead also, to make it clear this is the raw value.

tfry-git:
Note that we can actually have both (which will "look" like a single template function with two optional parameters). Yours definitely makes sense, as it allows full control. Mine, I feel, may also be justified, as it simplifies the (important?) use case of converting from Oscil/Sample/filters, even if it hides away the bit-counting.

Imagine, also, we had a ```template<int NF> AudioOutput_t MonoOutput::fromFraction(SFixMath<0, NF>);``` to go with that. This might then actually open entirely new use-cases. E.g.:

```
int8_t a = aSin.next();
int8_t b = kLFO.next();
int8_t c = master_volume;

//return MonoOutput::fromNbit(24, a*b*c);                       // Oops on AVR: Auto-promotion only goes up to 16 bit ints
return MonoOutput::fromFraction(toSFix(a)*toSFix(b)*toSFix(c));  // Let Mozzi handle the bit-counting
```

One step further, imagine the potential of ```toSFix(Oscil.next())```, when in a future version of Mozzi, Oscils will not be limited to 8 bits.

(Yes, this also has the potential to run into even 64bit calculations, unexpectedly, but then, this has always been an
area full of pitfalls, anyway.)

> I think that's a way to go, especially because the `::` is not encountered very often (and never in Mozzi), so I am a bit afraid of the troubles in can bring to users. Maybe `toSFixRaw` instead also, to make it clear this is the raw value.

Yes, the naming should be clear (but if possible not too cumbersome). "toSFraction" came to my mind. This should remove any expectation that a whole number is returned. OTOH it would also remove the obvious association with SFixMath, so not sure, whether that's a good idea, overall.

tomcombriat:
Very good points, basically we could use this kind of approach everywhere when a non-overflowing integer is useful. The autosizing is a good point, especially as Mozzi's own object do use the whole range of the integers they are based on. Powerful and dangerous tool…

--- 

I do not have lot of time these days but will try to put that together, I am just concerned (but without trying) that the compiler might complain that the templates should be clearer, even if all arguments are actually known, so that:

```
auto toto = toSFraction(aSin.next());
```
 would need to be
```
auto toto = toSFraction<>(aSin.next());
```

but I haven't tried.

----

`toSFraction` sounds good to me. Of course, a similar helper `toSInt` would also be of the party I think. I also think that these names would be misleading for cases which are mixed like:
```
auto tata = toSFix<5>(aSin.next()); // SFixMath<5, 2>
```
where I think `toSFix` sounds non-misleading (where `toSFraction<5>` does to me: it has integer bits in the end, it is not a fraction). This is probably the trickier one to make easy to understand for users, but I am actually not sure it is actually useful… Might try to leave it out at first, and see what it gives!

For `audioOutput` this is also a very good idea, I would push even further with a `return MonoOutput::fromSFix` (or `fromFix`), as basically, the only thing Mozzi needs to know, is `NI+NF`, which would allow the case (from your example) `a+b+c` to be handle in an exact same way. In combination with Oscils, as you suggested, this seems very powerful, and `toSFraction` will be more useful there than `toSInt` as the range will effectively be the same for all oscillators, just with different precision.

tomcombriat:
Okay, I added `.toUFraction()`, `toUInt()` and the same for `SFix`. Also made a general inverse and simplified the current implementation of `.invAccurate()`.

Now, something bugs me, but I am not sure there is an easy solution (but maybe ?). Let's us take the following example:

```
UFixMath<8,0> a, b, c, d;

auto toto = a+b+c+d; // leads to UFixMath<11,0> even though only 10 bits are actually needed
auto tata = (a+b) + (c+d); // leads correctly to UFixMath<10,0>
```

The problem here is that, to ensure non-overflow, promotion to the next bigger number of bits is needed, but adding two different types actually lets an *headroom*: `UFixMath<9,0> + UFixMath<8,0>` does not fill the resulting `UFixMath<10,0>` completely, a value of up to 255 (8 bits = (9-8)*8) can still can into that.
Now, numbers could remember the headroom they have left, but I do not see how that could be applied to change the returning type of say, the `+` operator between two different FixMath, even thought the headroom is known at compile time. Any idea @tfry-git ? Or maybe I am overthinking it, and just good documentation about parenthesing is needed.

For reference, here is the current addition operator:
```
  template<byte _NI, byte _NF>
  UFixMath<MAX(NI,_NI)+1,MAX(NF,_NF)> operator+ (const UFixMath<_NI,_NF>& op) const
  {
    constexpr byte new_NI = MAX(NI, _NI) + 1;
    constexpr byte new_NF = MAX(NF, _NF);
    typedef typename IntegerType<MozziPrivate::uBitsToBytes(new_NI+new_NF)>::unsigned_type return_type;
    UFixMath<new_NI,new_NF> left(*this);
    UFixMath<new_NI,new_NF> right(op);

    return_type tt = return_type(left.asRaw()) + right.asRaw();
    return UFixMath<new_NI,new_NF>(tt,true);
  }
```
would need to have its return type depending on the headroom remaining in the biggest of the two initial types. Would probably another big knot to implement, if possible…

tfry-git:
> Now, numbers could remember the headroom they have left, but I do not see how that could be applied to change the returning type of say, the + operator between two different FixMath, even thought the headroom is known at compile time.

I think the only way to do that would be as third template parameter (which could, obviously, default to 0 if not specified).

As for the follow-up question, I am not sure. My initial feeling was that you may be overthinking it, indeed, considering it will probably be non-typical to add a large amount of numbers, and the drawbacks of one bit too many, here or there, did not seem to bad. OTOH, considering the vision that SFixMath might actually provide a full solution to bit-counting inside updateAudio(), we'd really want to be as accurate as possible.

I'm a bit undecided, here. How much pain would it be to carry a headroom parameter through all operators? And how much confusion will it cause?

Thinking about alternatives, I believe the following additions may be useful in their own right (sorry for adding ever more complexity to this PR):
  - .trim(byte NI, byte NF) -> returns a SFixMath/UFixMath trimmed to the given dimensions, if necessary also shifting into the new range
  - .clip(byte NI, byte NF) -> same, but instead of shifting into the new range, only clip values in excess of the new range
  - .assert(byte minbits, byte maxbits) -> raise static_assert, if this number has more or less bits than expected
 (Maybe the above should also use template parameters, instead, and the .assert() may - or may - not, even have to).

In your example: ```auto tata = (a+b+c+d).clip(10,0);```.

tomcombriat:
> I think the only way to do that would be as third template parameter (which could, obviously, default to 0 if not specified).

Yes, this is also the conclusion I came to. Fortunately, only `NI` is impacted, not `NF`.

> I'm a bit undecided, here. How much pain would it be to carry a headroom parameter through all operators? And how much confusion will it cause?

Honestly, I do not pronounce myself on pain involved anymore… I started this thinking that would be a quick job, but it turns out that it is not that simple IMO, especially correctly figuring out the maths. A problem is that, there is also not a lot on that around, some people probably figured all of this out… 40 years ago. Anyway, I will try to see, maybe in another branch, and see how it goes. Regarding confusion, my idea would be to *completely hide* that to the user (but with documentation for the developers), in order to make this as carefree as possible.

The functions you propose could also be useful but as a clarification, would they *actually* check the value before trimming or clipping? Or would a complete templated approach, working on ranges, be what you have in mind? Note that clip can be done already with casting: `UFixMath<10,0> toto = a+b+c+d;` (except when there is an overflow maybe) , and trim with `.sR<>()` and `.sL<>()`.

I will try to make my mind about it… Will continue with the examples first as they usually bring some good usage cases.

> sorry for adding ever more complexity to this PR)

Sorry that it takes that long! I did not realized at first that this would turn out that big! But I also quite like the way this module behaves now. Thanks for taking the time to answer my questionings!

tfry-git:
> The functions you propose could also be useful but as a clarification, would they _actually_ check the value before trimming or clipping?

TBH, I may not have quite thought this through to the end, and in consequence it may be a good idea to ignore this bit, for the time being. It falls into the category of "can be added at any time", after all.

And yes, I had also severely underestimated the complexity of this PR. But then that's because I also underestimated the complexity of the problems that it solves. This is going to be good!

tomcombriat:
I think this looks quite good. There might be some possible improvements on the sides of `constexpr` and automatic testing but I think this is already a big baby.

Outside of the code itself, and before merging, I would like to gather your opinion on this: I think this library could also be quite useful in other context than Mozzi (at least, I am sure I would use it for other purposes). I think there are basically three options:

 1. Keep it here, noting that it can still be used without Mozzi.
   - Pros: easier
 2. Make it a separate library, that would be required when installing Mozzi (it is very likely possible to do that when Mozzi is installed via the Arduino Library Manager, for forks a clear indication should be provided).
   - Pros: it might attracts more users on this part of the code which in turn would provide more feedback to improve it and detects potential bugs. Could also be useful for other projects.
   - Cons: this adds a dependency to Mozzi
  3. A mixed solution where another version is given as a standalone library (with different names to avoid conflicts).
    - Cons: probably a nightmare to keep updated

I am not in favor for 3. because of the difficulties to maintain such a duplicated code, but I would be quite up to try to set 2. up, as a separate repo with you as collaborator(s). As this was developed in the context of Mozzi, I of course would like your opinions about it and the decision is taken together.

FixMath.h
```diff
@@ -0,0 +1,1287 @@
...
+  
+  /** Returns the internal integer value
+      @return the internal value
+  */
+  internal_type asRaw() const { return internal_value; }
+  
+  /** The number of bits used to encode the integral part.
+      @return The number of bits used to encode the integral part.
+  */
+  int8_t getNI(){return NI;}
```
tfry-git, line 494: Should be declared const, too (and constexpr, but let's delay worrying about constexpr to after this is merged).

FixMath.h
```diff
@@ -0,0 +1,1287 @@
...
+    @return A UFixMath<0,NF> with NF chosen according to the input type
+*/
+template<typename T>
+inline UFixMath<0, sizeof(T)*8> toUFraction(T val) {
+  return UFixMath<0, sizeof(T)*8>::fromRaw(val); 
+}
+
+/** Create a *pure* integer unsigned fixed number (UFixMath) from an unsigned integer.
+    The number of fractional bits (NI) is chosen automatically depending on the input 
+    type. Hence toUInt(255) and toSInt(uint8_t(255)) *do not* lead to the
```
tfry-git, line 622: toSInt() -> toUInt()

FixMath.h
```diff
@@ -0,0 +1,1287 @@
...
+template<typename T>
+inline UFixMath<sizeof(T)*8-1,0> toUInt(T val) {
+  return UFixMath<sizeof(T)*8-1,0>::fromRaw(val); 
+}
+
+
+
+/** Instanciate an signed fixed point math number.
+    @param NI The number of bits encoding the integer part. The integral part can range into [-2^NI, 2^NI -1]
+    @param NF The number of bits encoding the fractional part
```
tfry-git, line 639: (Also for UFixMath): AFAIU, RANGE is still "visible" (I for one don't know a way to hide it). Maybe it makes sense to document it as "purely internal"?

FixMath.h
```diff
@@ -0,0 +1,1287 @@
...
+
+    Under the hood, these types will keep track of the maximum possible value they might hold (this is the RANGE template parameter), and, if only *SAFE* operations (see below) are used, will automatically adjust there NI and NF to accomodate the result of a operation. It will also try not to promote there internal type when possible, assuming that you use the complete range of a given type.
+
+    The operations possible with these types can be divided into two categories:
+    - the operations between FixMath types are all safe (aka won't overflow) and are the only one included by default
+    - the operations between a FixMath and a native C type (int, float) are NOT safe and are not included by default. In order to activate them, you need to `#define FIXMATHUNSAFE` before including FixMath.h.
```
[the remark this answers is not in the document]
tfry-git: To keep doxygen happy, I believe it will also be necessary to add
```
#if FOR_DOYGEN_ONLY
#define MOZZI_FIXMATH_UNSAFE
#endif
```
at the top of this file.

FixMath.h
```diff
@@ -0,0 +1,1287 @@
...
+  
+  /** Returns the internal integer value
+      @return the internal value
+  */
+  internal_type asRaw() const { return internal_value; }
+
+  /** The number of bits used to encode the integral part.
+      @return The number of bits used to encode the integral part.
+  */
+  int8_t getNI(){return NI;}
```
tfry-git, line 986: const. See above.

Oscil.h
```diff
@@ -150,6 +151,19 @@ class Oscil
...
 
 
+	/** Returns the next sample given a phase modulation value.
+	@param phmod_proportion a phase modulation value given as a proportion of the wave. The
+	phmod_proportion parameter is a SFixMath<15,16> fixed-point number where the fractional part represents almost -1 to almost 1, modulating the phase by one whole table length in
+	each direction.
+	@return a sample from the table.
+	*/
+	inline
+	int8_t phMod(SFixMath<15,16> phmod_proportion)
```
tfry-git, line 161: Wouldn't this be much more powerful as a template function taking *any* SFixMath?

Oscil.h
```diff
@@ -203,6 +217,19 @@ class Oscil
...
 
+  	/** Set the frequency using UFixMath<24,8> fixed-point number format.
+	This might be faster than the float version for setting low frequencies such as
+	1.5 Hz, or other values which may not work well with your table size. A UFixMath<24,8>
+	representation of 1.5 is 384 (ie. 1.5 * 256). Can't be used with UPDATE_RATE
+	less than 64 Hz.
+	@param frequency in UFixMath<24,8> fixed-point number format.
+	*/
+        inline
+	void setFreq(UFixMath<24,8> frequency)
```
tfry-git, line 228: ~~See above. A template function would be much more powerful.~~

Nevermind. However, I suggest moving the template-version of setFreq(), upwards, and declaring this one as a template-specialization, explicitly.

Oscil.h
```diff
@@ -226,6 +253,49 @@ class Oscil
...
+	/** Set the frequency using UFixMath<16,16> fixed-point number format. This is useful in
+	combination with Q16n16_mtof(), a fast alternative to mtof(), using UFixMath<16,16>
+	fixed-point format instead of fractional numbers.
+	@note This should work OK with tables 2048 cells or smaller and
+	frequencies up to 4096 Hz.  Can't be used with UPDATE_RATE less than 64 Hz.
+	@note This didn't run faster than float last time it was tested, after 2014 code changes.  Need to see if 2014 changes improved or worsened performance.
+	@param frequency in UFixMath<16,16> fixed-point number format.
+	*/
+        inline
+	void setFreq(UFixMath<16,16> frequency)
```
tfry-git, line 267: See above.

examples/01.Basics/Vibrato/Vibrato.ino
```diff
@@ -41,7 +45,8 @@ void updateControl(){
 
 
 AudioOutput_t updateAudio(){
-    Q15n16 vibrato = (Q15n16) intensity * aVibrato.next();
+  //Q15n16 vibrato = (Q15n16) intensity * aVibrato.next();
+  auto vibrato = intensity * SFixMath<0, 7>::fromRaw(aVibrato.next()); // Oscils in Mozzi are 8bits: 7bits of signal plus one bit of sign, so what they fit into a SFixMath<0,7> which is a signed fixMath type with 7 bits of value.
```
tfry-git, line 49: Use toSFraction(), or at least mention it as an alternative.

examples/01.Basics/Vibrato_Midi_Note/Vibrato_Midi_Note.ino
```diff
@@ -0,0 +1,64 @@
...
+
+void setup() {
+  startMozzi(CONTROL_RATE);
+  aCos.setFreq(mtof(midi_note));
+  kVibrato.setFreq(UFixMath<16, 16>(10));  // frequency of the modulation
+}
+
+
+void updateControl() {
+  auto modulation = SFixMath<0, 7>::fromRaw(kVibrato.next()) * mod_amplitude;  // Oscil in Mozzi are fundamentally 8 bits: 7bits of data and 1bit of sign.
```
tfry-git, line 50: See above. toSFraction()

tfry-git (commented):
I did not look at everything super closely, this time around, but looks good to me overall, with the usual nitpicks.

Two slightly larger matters:
- I still hope to implement a testing-scheme based on static_asserts(), but as discussed, this will need additional work on constexpr. But I also suggest to ignore that until after this is merged. This PR is complex enough, already.
- Would it make sense to rename the classes as UFix~~Math~~/SFix~~Math~~?

Oscil.h
tomcombriat: Indeed, like the way I did for `setFreq()` you mean? Will do.

examples/01.Basics/Vibrato/Vibrato.ino
tomcombriat: Yes, I did not do the examples, planning for maybe a consecutive PR, but as these two examples are in this one this should be coherent.

FixMath.h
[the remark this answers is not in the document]
tomcombriat: done.

FixMath.h
```diff
@@ -0,0 +1,1287 @@
...
+      - UFixMath<NI,NF> + UFixMath<_NI,_NF> returns UFixMath<MAX(NI,_NI)+1,MAX(NF,_NF)> at worse
+      - SFixMath<NI,NF> + SFixMath<_NI,_NF> returns SFixMath<MAX(NI,_NI)+1,MAX(NF,_NF)> at worse
+      - UFixMath<NI,NF> + SFixMath<_NI,_NF> returns SFixMath<MAX(NI,_NI)+1,MAX(NF,_NF)> at worse
+      - UFixMath<NI,NF> + anything_else returns UFixMath<NI,NF> (only available with `FIXMATHUNSAFE`)
+      - SFixMath<NI,NF> + anything_else returns SFixMath<NI,NF> (only available with `FIXMATHUNSAFE`)
+    - Subtractions:
+      - UFixMath<NI,NF> - UFixMath<_NI,_NF> returns SFixMath<MAX(NI,_NI),MAX(NF,_NF)> at worse
+      - SFixMath<NI,NF> - SFixMath<_NI,_NF> returns SFixMath<MAX(NI,_NI)+1,MAX(NF,_NF)> at worse
+      - SFixMath<NI,NF> - UFixMath<_NI,_NF> returns SFixMath<MAX(NI,_NI)+1,MAX(NF,_NF)> at worse
+      - UFixMath<NI,NF> - anything_else returns UFixMath<NI,NF> (only available with `FIXMATHUNSAFE`)
```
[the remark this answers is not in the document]
tomcombriat: Done.

FixMath.h
tomcombriat: Good catch, and a few others too!

Oscil.h
tomcombriat: Done (I think?)

examples/01.Basics/Vibrato_Midi_Note/Vibrato_Midi_Note.ino
tomcombriat: Done.

Oscil.h
tomcombriat: Done

tomcombriat:
Thanks for your comments @tfry-git !

> Would it make sense to rename the classes as UFix~Math~/SFix~Math~?

Indeed, I think it looks better. This is done.

---

> Outside of the code itself, and before merging, I would like to gather your opinion on this: I think this library could also be quite useful in other context than Mozzi (at least, I am sure I would use it for other purposes). 

@tfry-git , @sensorium any opinions on this?

tfry-git:
> > Outside of the code itself, and before merging, I would like to gather your opinion on this: I think this library could also be quite useful in other context than Mozzi (at least, I am sure I would use it for other purposes).
> 
> @tfry-git , @sensorium any opinions on this?

As discussed in private mail, I do think that makes sense, indeed. I'll add that I'm not sure, what will be the best procedure:

I *imagine* that Mozzi is not a bad breeding ground for the time being (as it includes auto-compilation infrastructure and a "playground" of examples to test the FixMath concepts on). So, in other words, I think it may (or may not) make sense to delay the "split-out" for another little while. What's your plan on this?

tomcombriat:
> As discussed in private mail, I do think that makes sense, indeed. I'll add that I'm not sure, what will be the best procedure

Honestly I do not really know, especially how the automatic testing will coop with a dependency. If I am correct, the Arduino IDE does fare well with duplicated libraries hence my plan for now was to start another repo (probably today), and get it into the library manager while continuing to work on this PR (`Line` requires a bit of work, as well as finishing the examples). It will become more gigantic but t least it will be on one thing.
 Once the other one is on, we can try to remove the source file for FixMath here and take it from there? What do you think?

tomcombriat:
I would be in favor of merging this into Mozzi2 if we are satisfied with the current implementation (which could change). 

There is still work to do, including changes in some classes (`Line`, `Smooth`) which might imply some design choices, deserving their own PR. Also all change on example is now going to lead to a conflict (I could merge Mozzi2 in here though, but I have the feeling this PR is big enough).

That does not preclude the design of FixMath to change of course.

PR #215, opened 2023-11-29
Fix of mozziAnalogRead for Renesas

tomcombriat:
To keep the git on track, I am duplicating the PR on master here. I hope that's the way to go!


tomcombriat:
Merging. Will revert if adverse consequences should arise.

PR #222, opened 2023-12-24
mozzi_midi as header only

tomcombriat:
I did not try very hard to make the table private (gave up at the first error), but trying with several files importing it does not cause any problem. Will try more intensively if deemed necessary!

Best


tomcombriat:
In the end, I made the table private. Reading a bit more about static variables in headers, it seems fairly clear that this table might end up duplicated otherwise. Good on my side!

tfry-git:
Looks good to me. It also seems to enable some massive flash savings in some scenarios (probably, only where the compiler can infer which midi notes are used though?).

tomcombriat:
> It also seems to enable some massive flash savings in some scenarios

Ah yes, indeed, I only noticed the RAM… My take would be that this big table is not allocated at all for all examples which are not using `mozzi_midi.h`, whereas before it was compiled and allocated all the time?

tfry-git:
Hm, experimenting some more, it turns out, when including mozzi_midi.h from more than one .cpp-file (in the user sketch), like this, we will be violating the "one definition rule". Contrary to class member functions, global functions may be defined only once, even if their definitions match. This restriction can be lifted by declaring these functions (and the midiToFreq-table, too) "inline". That in turn is generally ignored as an optimization hint, these days (I tested, and avr-gcc ignores it, too), but lifts the strict one definition rule.

In short: Prefix all definitions with "inline".

---

As a side-note, I believe to have understood the reason why AVR (contrary to the other platforms) does not show these massive flash improvements: It's because the table is forced into PROGMEM-space, there, and the compiler is not able to deduce that the full table isn't still needed there. (Not something we can do anything about, I believe, and also unlikely to matter in real world scenarios.)

tomcombriat:
> Hm, experimenting some more, it turns out, when including mozzi_midi.h from more than one .cpp-file (in the user sketch), like this, we will be violating the "one definition rule". 

Ah, good that you figured out because I had the feeling it worked when I tested (but I might have included in several header, not source file). Will do!

tomcombriat:
Okay, indeed I have the same problem when I include `mozzi_midi.h` in another user source file. I guess this also concerns FixMath2.0 (#212).

> and the midiToFreq-table, too

Unfortunately, it seems that C++-17 is needed for that (complain from the compiler: "inline variables are only available with '-std=c++17' or '-std=gnu++17". It also creates a conflict with the Oscillators tables that I did not investigated…

My current solution is to move the table to its own source file, but maybe you have a better solution? This kills a bit the idea that everything in the header but at least all functions definitions are at the same place.

tomcombriat:
> guess this also concerns FixMath2.0 (https://github.com/sensorium/Mozzi/pull/212).

Actually it does not… But I do not really why. I know that the functions defined inside the class are considered `inline` by default, hence not affected, but all the reverse overloads should be from my understanding…

tfry-git:
I must admit, I'm still learning about the one definition rule, too, while working on the "single compilation unit" thing. Fortunately, no true road block(*) in sight, but a bunch of small complications that I had not foreseen.

Anyway, I *believe*, what may work is this:

class MozziMidiPrivate {
private:
  friend function int mtof(uint8_t); // add other functions using this, here

  static uint32_t miditoFreq(uint8_t which) {
    static CONSTTABLE_STORAGE(uint32_t) lot[128] = { ... };
    return FLASH_OR_RAM_READ<const uint32_t>(lot + which);
  }
};

I.e. move the definition into a function body, instead, where it ought to be acceptable.

---

(*) Besides, strictly speaking, we won't really be needing a *single* compilation unit, but only a single compilation unit that is affected by configuration defines.

tomcombriat:
Okay, the compiler seems to be happy this trick! I just have to test on some hardware and do a bit of cleanup.

Actually, as you said, some things will probably do not need to be in the "main" translation unit. Hence I wonder if keeping the table out in its own cpp file is actually cleaner in order to keep that header clean?

tfry-git:
Learning a new error message every day... This is what the internet has to say about it: https://stackoverflow.com/questions/35091862/inline-static-data-causes-a-section-type-conflict . An interesting read, but to me it boils down to: Too difficult to bother with. So, yes, let's indeed keep the table in a separate .cpp file, then (while waiting for availability of C++17).

tfry-git (approved):
Looks good to me. Perhaps add a header to mozzi_midi_table.cpp?

New theory on the flash savings: The compiler optimizes away the floating point computations where mtof() is called with a compile-time constant. This may result in needing no support for floating point arithmetic at all -> massive savings.

tomcombriat:
Merging.

Issue #210, opened 2023-11-13
RFC: Mozzi 2.0 part 3 - Miscellanious


tfry-git:
Here's just a list of smallish stuff that should probably be changed for a 2.0. Add your items in comments (assuming github won't let you edit this summary), and speak out if you disagree with any.

- [x] Instead of ```#include <MozziGuts.h>```, let's make it ```#include <Mozzi.h>```. The former could still be supported (with a warning), and in fact that could help detect sketches that still assume Mozzi 1.x. (possibly enable extra compatibility workarounds).
- ~~[ ] Remove ```control_rate_hz``` parameter from ```startMozzi()```. That's controllable via ```CONTROL_RATE```, already, and merely introduces ambiguity (plus a few unneeded bytes). (It made sense before config restructuring in Mozzi 2.0, in that for this particular value, setting it at the top of the sketch was actually possible).~~ Does not actually save any flash.
- [x] Move ```CONTROL_RATE``` define up in sketches, for consistency with other config options.
- [ ] Quite a bunch of (purely internal) #defines do not actually control code compilation, but are merely constants. It should be possible to replace these with ```const```s without affecting code size. The advantage is that code will be cleaner, and the defines can then be hidden in a private namespace, for instance. Also they'll be implicitly documented as not user-modifiable. Of course, we should monitor for any unexpected code-size changes carefully, when doing this, so best to handle this as a separate step. (One of these is MOZZI_AUDIO_BIAS) -> not so easy to accomplish
- [ ] There's a lot of duplicated information e.g. between README.md and API (but also within the API), and typically one or the other (or both) is partially outdated, already. We should try to avoid that. In particular, I think the specific hardware notes & config options will be most likely to remain up to date, when kept inside the API. The README should have direct links to that documentation, instead. -> in progress
- [x] Add auto-builds for more boards: Teensy and Renesas should be possible.
- [ ] Add auto-builds for various configs. -> Globally, audio input is not yet compiled. Regarding port-specific options, I have no clear idea on how to integrate that into those into the auto-builds.
- [x] Move purely internal header files to "internal" (including all the MozziGuts implementations)
- [ ] Update "learn" pages on gh-pages! -> in progress
- [x] Remove obsoleted "STEREO_HACK"
- [x] use of integer/void return value of updateAudio() should be disabled according to compatibility level
- [x] Automatic scaling for analog (and audio) input
[...]

PR #224, opened 2024-01-04
Single compilation unit / In-line configuration

tfry-git:
Notes:
- PR was created early to monitor compilation statistics while working, but this is now approaching a useful state.
- [x] No compilation testing done for Teensy
- We now have some true (compilation) testing coverage of MOZZI_OUTPUT_EXTERNAL_*, MOZZI_OUTPUT_2PIN_PWM, and MOZZI_STEREO, *but* on platforms that don't support the latter two config options, there is a hack in place to silently ignore those - if and only if running inside the github workflow.
- Changes certainly look massive, although to a large degree bits have simply been moved around, unchanged. I'll try to write up some hints on what to look out for in the coming days.

Regarding flash (and RAM) usage changes:
I left only the reports at a few steps that I think are meaningful, to get a clearer picture, and removed the other ones. What we see is:
1. First some minor savings across the board after moving trivial bits around (makes sense, as compiler can now avoid a few function calls).
2. After the main chunk (inlining MozziGuts.cpp), most platforms show moderate savings, but AVR shows +60 bytes flash usage on a few selected sketches. I have no real idea, why this is happening. Comparing the disassembly side-by-side, It does not even appear to have anything (recognizable) to do with the changes contained in this PR, rather some Oscil code is apparently getting inlined, for whatever reason.
3. Removal of TimerZero.cpp (only marginally related to this PR) papers over this effect, by saving 98 bytes of flash on AVR.
4. No changes during some more formal (documentation related, etc.) changes.
5. After inlining the config options in the examples, some platforms (importantly, those supporting Stereo) show a more diverse picture again, which is not actually a surprise, however: Now these examples are finally actually compiled with the intended configuration.


tfry-git:
I promised to explain a bit, what I have done (not in any particular order).

1. Moved MozziGuts_impl_XYZ headers (and some others) into the "internal" dir.
2. Moved a few hardware-specific bits out of mozzi_analog.h (into MozziGuts_impl_AVR.hpp).
3. Moved the random seeding code out of mozzi_rand.cpp and into MozziGuts_impl_XYZ.hpp. This required some acrobatics to comply with the one definition rule, too, which is why part of the implementation now ended up in a (private) class MozziRandomPrivate. In this case, fortunately, no mysterious compiler errors, because no section attributes need to be specified.
4. Removed (AVR-only) TimerZero code which had been unused for years. Interestingly, this was actually still clogging up flash space due to the ISR definition contained in the lib (compiler cannot know that it wasn't used, somewhere).
5. Modified MozziGuts.h so it includes MozziGuts.hpp (formerly MozziGuts.cpp) into the same translation unit. That's actually the central (and only strictly required) step (as the other .cpp files do not make direct use of configuration defines). Now this means that a whole lot of internal are actually in the same translation unit as the user sketch. To avoid catastrophic clashes, here, I introduced a namespace MozziPrivate, and put almost everything into that. This latter was actually the most tricky part, because certain things need to stay at the global scope, and others not. I made a few mistakes there, at first, but hope to have ironed those out. The Teensys still need (compilation) testing, however.
6. Introduced two "wrapper" includes: Mozzi.h and MozziHeadersOnly.h. Mozzi.h will have to be present exactly once in each project from now on, while MozziHeadersOnly.h is for cases, where Mozzi core declarations are needed in more than one translation unit. (Continuing to include MozziGuts.h is still ok, as only one translation unit is involved, but triggers a message, that you should probably update some stuff).
7. Adjusted the examples, accordingly. Not sure, whether I've caught everything. Keep in mind that our automatic compilation covers only a fraction of the examples, so far. Importantly however, stereo and hifi examples are now actually using an appropriate config (specified inline). So do the external output examples, but those are not yet part of the workflow runs. Platforms that don't support stereo/hifi have a hack to ignore that setting, if - and only if - compiling inside the github workflow.
8. Fought hard with Doxygen to not just throw away the documentation for all the defines, and ignore half of the documentation groups. (Actual update of generated API-documentation is not contained in this PR).

TLDR:
- Could you test the Teensies once more? And perhaps an external output sketch or two?
- Comments on the MozziPrivate namespace hack?
- D'accord on switching to Mozzi.h as the new main include file?

tomcombriat:
Without looking at the diffs yet (I tried to kept up to date, but that went a bit fast in the end…), some preliminary answers:

> D'accord on switching to Mozzi.h as the new main include file?

Completely, this actually makes more sense to me, name-wise.

> Comments on the MozziPrivate namespace hack?

Seems like a clean solution to me, isn't that what namespaces are for basically?

> Could you test the Teensies once more? And perhaps an external output sketch or two?

Will try to do soon, alongside a more in-depth review!

Mozzi.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Mozzi.h
+ *
+ * Copyright 2023, Thomas Combriat and the Mozzi team
```
tomcombriat, line 4: Should probably be edited with the correct surname ;)

MozziHeadersOnly.h
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Mozzi is licensed under a Creative Commons
+ * Attribution-NonCommercial-ShareAlike 4.0 International License.
+ *
+ */
+
+/** @ingroup core 
+ * @file MozziHeadersOnly.h
+ *
+ * This file provides declarations of the \ref core Mozzi functions, but no implementation. Use this only, if you have more than one
```
tomcombriat, line 16: So the idea is that Mozzi.h should always be included in the main sketch and MozziHeaderOnly.h in every user cpp file?

Readme_Mozzi_2_0.md
```diff
@@ -28,16 +28,17 @@ all new
...
-you should avoid these functions, as they are slow (or even blocking). For measuring time, refer
-to mozziMircos(). For delaying events, you can use Mozzi's EventDelay() unit instead
-(not to be confused with AudioDelay()).
+ - Teensy3/4: channel2sc1a -> thought to be unused, removed
+ - Teensy2: adc_mapping -> hidden away; use adcPinToChannelNum(), as on all other platforms, instead
+ - removed several inclusions of "WProgram.h". If using Arduino versions < 1.0, you need to update, seriously ;-)  (TODO many, many, instances of this are still around)
+ - Since Mozzi (AVR-port) no longer uses Timer 0 since a long time, the corresponding library (utility/TimerZero.h) has now been removed, too.
+   The Arduino functions delay(), millis(), micros() and delayMicroseconds() should now be usable in theory. That said,
+   you should avoid these functions, as they are slow (or even blocking). For measuring time, refer
+   to mozziMircos(). For delaying events, you can use Mozzi's EventDelay() unit instead (not to be confused with AudioDelay()).
```
tomcombriat, line 40: Small typo (`mozziMicros()`)

examples/01.Basics/Sinewave_HIFI/Sinewave_HIFI.ino
```diff
@@ -41,15 +35,18 @@
     Tim Barrass 2012-13, CC by-nc-sa.
 */
 
-#include <MozziGuts.h>
+#include <MozziConfigValues.h>
+#define MOZZI_AUDIO_MODE MOZZI_OUTPUT_2PIN_PWM
```
tomcombriat, line 39: Clean! 
I wonder if, in the future, it might be worth to give access to some configurations with several defines, straight with a macro (I am thinking of something close to the `MIDI_CREATE_INSTANCE` from the Midi library for instance.

examples/01.Basics/Skeleton_Multi/Skeleton_Multi_Unit2.cpp
```diff
@@ -1,4 +1,6 @@
-#include <MozziGuts.h>        // This file, too, will have to include the Mozzi headers.
+#include <MozziHeadersOnly.h>        // <Mozzi.h> should be included only once in the whole program. Sketches needing
+                                     // core Mozzi functions in more than one .cpp file, shall include MozziHeadersOnly.h
+                                     // in all but one.
```
tomcombriat, line 3: That answers my previous question. I think this deserves a comment in the Learn of FAQ of the website at some point.

examples/04.Audio_Input/Audio_Input/Audio_Input.ino
```diff
@@ -1,11 +1,10 @@
 /*  Test of audio input using Mozzi sonification library.
 
- An audio input using the range between 0 to 5V on analog pin A0
- is sampled and output on digital pin 9.
+ An audio input using the range between 0 to 5V on analog pin A0 (or as
```
tomcombriat, line 3: Detail, but now there is quite some support for 3.3V boards, I would just change that to make the 5V less absolute (not related to the PR)

examples/12.Misc/Stereo/Stereo.ino
```diff
@@ -12,7 +14,12 @@
  *  This example code is in the public domain.
  */
 
-#include <MozziGuts.h>
+// Configure Mozzi for Stereo output. This must be done before #include <Mozzi.h>
+// MozziConfigValues.h provides named defines for available config options.
+#include <MozziConfigValues.h>
+#define MOZZI_AUDIO_CHANNELS MOZZI_STEREO
```
tomcombriat, line 20: Wonder if (that might already be the case), `#define MOZZI_AUDIO_CHANNELS 2` also work?

MozziHeadersOnly.h
tfry-git: At least in every .cpp file that needs Mozzi delcarations, yes. I added a minimal example to that effect as Basics->Skeleton_Multi

internal/MozziGuts.hpp
```diff
@@ -266,19 +269,51 @@ unsigned long mozziMicros() { return audioTicks() * MICROS_PER_AUDIO_TICK; }
...
 
 // reduce Macro leakage
 #undef LOOP_YIELD
 #undef BYPASS_MOZZI_OUTPUT_BUFFER
 #undef AUDIO_HOOK_HOOK
 #undef AUDIOTICK_ADJUSTMENT
 #undef MOZZI__LEGACY_AUDIO_INPUT_IMPL
+
+// "export" publicly accessible functions defined in this file
+// NOTE: unfortunately, we cannot just write using MozziPrivate::mozziMicros(), and that will conflict with, rather than define mozziMicros()
```
tomcombriat, line 295: I might be a bit slow, but for the sake of learning could you explain me why `MozziPrivate::startMozzi()` would not work for instance?

internal/MozziGuts_impl_TEENSY.hpp
```diff
@@ -42,14 +45,16 @@ void setupFastAnalogRead(int8_t speed) {
...
 {
-  advanceADCStep();
+  MozziPrivate::advanceADCStep();
 }
+namespace MozziPrivate {
 
 void setupMozziADC(int8_t speed) {
   adc = new ADC();
-  adc->adc0->enableInterrupts(adc0_isr);
+  adc->adc0->enableInterrupts(adc0_isr);  // TODO: is this even correct? Does the function take a parameter? And is adc0_isr a predefined name, or are we free to move this to MozziPrivate?
```
tomcombriat, line 57: Not sure if I understood your comment correctly, but `adc0_isr` is not a predefined name, we define this function just above. Knowing if the isr can be moved in  `MozziPrivate`, I don't see why not.

tomcombriat: Or maybe not, now seeing your comment on the template.

internal/disable_2pinmode_on_github_workflow.h
tomcombriat, line 1: Neat!

tomcombriat (commented):
Okay, I scanned through and left some minor comments here and there, principally to get a good grasp of what is going on.
Will try to do some testing soon!

tomcombriat:
## Testing

I will update here what tests I do and there results as I go.

| Board | Condition | Compiles | Works | Comment |
| ---------| ---------------|---------------|-----------|----------------|
| AVR |EXTERNAL PT8211 |yes |yes | straight from the example|

tomcombriat (approved):

Mozzi.h
tfry-git: Will do ;-). Although I'll change the first name too, claiming Mozzi.h itself would seems a little over the top.

examples/01.Basics/Sinewave_HIFI/Sinewave_HIFI.ino
tfry-git: You mean e.g. setting 2pin-mode along with the pins to use in one macro?

examples/01.Basics/Skeleton_Multi/Skeleton_Multi_Unit2.cpp
tfry-git: Indeed. Adding a task item in the documentation PR, and "resolving" it, here.

examples/04.Audio_Input/Audio_Input/Audio_Input.ino
tfry-git: True. There are more facets to the problem, too. E.g. many examples give a specific pinout / circuit for Arduino Uno. Helpful as that is, it is no longer accurate in many cases. Not sure, what could be a clean and helpful way out, so I ignored that for now. Adding as a task item to #210 for the time being.

examples/12.Misc/Stereo/Stereo.ino
tfry-git: Indeed it would, MOZZI_STEREO is defined as 2.

internal/MozziGuts.hpp
tfry-git: In a vain attempt at keeping the source somewhat readable, I trying not to "enter" and "leave" the MozziPrivate-namespace too often. I.e. the basic idea was to write

```
namespace MozziPrivate {
  void privateFunction1() {}
  void privateFunction2() {}
  void publicFunction1() {}
  void privateFunction3() {}
}
void publicFuction1() { MozziPrivate::publicFunction1(); }
```

rather than:

```
namespace MozziPrivate {
  void privateFunction1() {}
}

namespace MozziPrivate {
  void privateFunction2() {}
}

void publicFunction1() {}

namespace MozziPrivate {
  void privateFunction3() {}
}
```

Maybe the second variant is preferable after all? I'm not sure at all (too bad we cannot just write ```void MozziPrivate::privateFunction1() { [definition] }```).

tfry-git: Or perhaps this (ignoring convetional code-formatting rules)?

```namespace MozziPrivate { void privateFunction1() {
  [definition]
}}```

tomcombriat: But in that case, users would need to use the Mozzi namespace right? Or would the "export" part would still be there?

A global namespace (`mozzi::`) would be neat for all functions in my opinion but:

 - it might be intimidating to some users
 - this would break existing sketches badly (we are already there, but not *that* badly).

tfry-git: Two somewhat separate but related topics, here:

As for the different variants of defining the MozziPrivate-namespace: The latter two variants would not need the "export" trick, because the "public" functions would not be placed inside MozziPrivate, in the first place. In contrast, the first variant (== current state of the PR) puts everything into MozziPrivate, and then "exports" the public ones.

I have indeed also been thinking a bit about adding a mozzi-namespace. Note that we could make it such that a default setup and/or examples could simply include a ```using namespace mozzi;``` line, which could again be turned off by some config switch, if desired. I have already added a MOZZI_COMPATIBILITY_LEVEL config define (mostly unused for now), which could also help with such a transition (while the compatibility level is set to < 2.0, the ```using``` line would be included by default, to keep old sketches working, but for >= 2.0, the default could be reversed).

In other words, the impact on users could be kept quite minimal, but without a doubt, we'd need to add a whole bunch of "namespace" statements, everywhere.

tfry-git:
Ok, in order to keep everything from diverging too far, I think I'll merge this (into devel/Mozzi2), later tonight. As far as I understand, the FixMath2 branch is currently stalled waiting on this, because you are still going to touch the examples, right? I'll try to get that in sync, too.

We can still address the question of how to best set up the MozziPrivate-namespace, later. (Even after Mozzi 2.0, as it won't be user visible).

Regarding your question about additional options in config_example_rp2040_i2s_pt8211.h, I concur, we should not make too much detail too visible. As a rule of thumb, I'd say, anything that isn't immediately expected to make a useful difference "outside the box" should be considered an implementation detail and remain hidden. (E.g. the number of separate buffers).

tomcombriat:
Hi,

>  As far as I understand, the FixMath2 branch is currently stalled waiting on this

Well, not really… It is stalled because I do not manage to put enough time on it these days… But it starts to come to an end, `Line` class adaptation, optimizations and examples change remain and that's it…

> We can still address the question of how to best set up the MozziPrivate-namespace, later. (Even after Mozzi 2.0, as it won't be user visible).

I agree. I did not have to work on that file yet (but that might come with the next port I have in mind for the ATTiny), so I honestly do not have a super clear idea of what is best…

PR #226, opened 2024-01-06
Draft: Mozzi 2.0 website updates

tfry-git:
So the other day we mentioned updating the docs. Here's a PR with a beginning regarding the "learn" section.

I set this up in my own fork, so we can have a "live demo" at https://tfry-git.github.io/Mozzi/ . Obviously the first commit, with change of title and color theme in _config.yml is just so it's easier to tell the two sites apart, and not meant to end up in the true website.

I'm not sure, whether you can technically commit to this PR, directly, otherwise I'll try to integrate comments as usual.

Misc TODO:
- [ ] Add FAQ / documentation on multiple definition errors / working with multiple translation units
- [ ] Add documentation / tutorial on customizing options


tomcombriat:
## Content proposition: 
(This comes from what I use for synthesizers powered by Mozzi)

---

## More Advanced Filter with Volume control 
LPF


Simple low-pass filtering is usually enough when outputting straight from chips without an DAC (Digital to Analog Converter), especially as they usually have a limited number of bit resolution (8 or the like).

However, more advanced outputting schemes, for instance with an audio DAC (usually 16bits), can greatly benefit more advanced filtering and power stabilization. This schematic is an example of such filtering. It implements a second order low-pass filter (which will attenuate more quickly frequencies beyond its cutoff than a first order) with stabilized audio-type operational amplifiers (TL074).

### Quick functional description
#### Stabilization and virtual mass
The top left part of the circuit provides a clean voltage source from the 5V available on most boards. This is critically needed especially if you are powering the board via USB which is extremely noisy. This filtered 5V is used to power the Op-Amps. While we are at it, we create a *virtual mass*, at 2.5V. This is needed because the TL074 is not a *rail-to-rail* amplifier: it cannot provide voltages close to its powering values (+5V and GND here), especially towards GND. As signals created both by a DAC or PWM always have GND as the minimal value, part of the signal would be clamped by the amplifier, resulting in saturation. In order to resolve that, we are going to biased the signal to 2.5V which means that our waveform will be centered on 2.5V instead of having 0V as a lower value.

#### Bias and gain
The first stage after the AUDIO IN is taking care of biasing the signal at 2.5V. The first two capacitors are first removing the DC component of the signal which is then centered on 0V, and the OP-AMP biased it to 2.5V.
This stage also comports a gain setting with the trimmer of 10k. The value of the trimmer should be adjusted at the end so that no saturation happens when sound is outputted by Mozzi.

#### Filtering
The next two OP-AMP are performing a second order low-pass-filter on the signal, removing high frequencies (DAC stepping, EM noise). It cutoff frequency can be changed by the value of one resistor between 10kHz (22kOhm) and 16kHz (10kOhm). 10kHz is usually the good choice.

#### Volume
The last stage implements a simple volume control with 10k potentiometer (linear). The 1k resistor attached to it gives the potentiometer a response which is closer to a logarithmic potentiometer which may be harder to find. Of course a logarithmic potentiometer (without the additional resistor) does perfectly the job.
This is followed by a simple buffer to bring down the output impedance. The output capacitors are there to remove the bias and give the next stage a signal centered on GND, which preferred in audio.

#### Note on the DC capacitors
DC capacitors should normally be of non-polarized type. However, it can be hard to find non-polarized capacitors with values in the range of the microFarad (or expensive). The circuit here uses the trick that a couple of polarized capacitors, with reversed polarities behaves like a non-polarized capacitor. Of course, these can be replaced with a single, non-polarized one.

tomcombriat:
Quick comment (sorry it comes after all the html edit…). Does Stereo really deserve a line on its own? The goal there is only to specify which pin the second channel use, or if it is not supported, maybe putting it after PWM-1 or PWM-2 would make sense? (After all, some of the DAC are stereo compatible, but still on the same line).

Sorry, I am stalking but it just popped in my notifications ;)

tfry-git:
> Quick comment (sorry it comes after all the html edit…). Does Stereo really deserve a line on its own? The goal there is only to specify which pin the second channel use, or if it is not supported, maybe putting it after PWM-1 or PWM-2 would make sense? (After all, some of the DAC are stereo compatible, but still on the same line).

I'll admit, I'm still not really happy with the table. The main goal is to show where audio goes in the default configuration (and I'm afraid, the formatting does not make this very visible), and a secondary goal was to give an overview of supported features. This may need a bit more experimentation to present in a useful way.
 
> Sorry, I am non stalking but it just popped in my notifications ;)

Hey, that's what this PR is for!

tfry-git:
> Hi All, admiring the work on all fronts (not just the docs). Would it be OK if I try to fix any typos in the draft docs if I notice them while reading through? What would the process be? Should I what till they're pulled into the main repo? Showing my lack of Git(hub) knowledge/etiquette ...

At any rate, it sounds very reasonable to fix all my typos and mistakes right away. This PR will be somewhat longer in the making, as it can only really be finalized along with Mozzi 2.0.

To be honest,though,  I'm not sure on the procedure, either. I *believe*, with the github command-line tool, it should be possible to do

```
gh pr checkout 226
# do your work
git commit -a
git push git@github.com:tfry-git/Mozzi.git:gh-pages
```

If that doesn't work out for some reason or another, I suppose I could move this work over to some branch inside the main repo (currently living in my personal fork, in order to test the rendering), which should ease collaboration. Should I?

tfry-git:
Closing this one in favor of #229 (same thing but within the main repo, for easier collaboration).

PR #230, opened 2024-01-27
Draft: Mozzi 2.0 miscellanious small stuff

tfry-git:
This PR will be some time in the making too. It's intended to test-compile "small" changes for Mozzi 2.0 (see also #210).

Feel free to add your own stuff, too. The changes here are not logically related except for begin targetted at Mozzi 2.0.

What's in this PR:

- Removed ```STEREO_HACK```. With Mozzi 2.0, any changes to mozzi_config.h are obsolete, anyway.
- Some cleanups in AudioOutput.h:
  - Replaced some defines with constexpr functions.
  - Obsoleted the separate ```AudioOutput_t``` that we had in addition to ```AudioOutput``` (it can still be used, but will show a warning). Under the hood, the in mono config, ```updateAudio()``` will now return ```MonoOutput```, rather than ```int```. (And still further under the hood, the compiler optimizes away everything. We're just removing a little weirdness in the code).
- Rename ```AUDIO_RATE``` and ```CONTROL_RATE``` to ```MOZZI_AUDIO_RATE``` and ```MOZZI_CONTROL_RATE```, everywhere. Since this could affect a larger number of sketches, the old names are still defined, and can still be set, but they simply aren't recommended, anymore, for consistency.
- Move ```MOZZI_CONTROL_RATE``` up in all examples. This is for consistency with other config options, removes a compile-time warning, but may also avoid some subtle bugs: E.g. Phasor.h makes use of the define, and changing it after including that would have unexpected effects.
- Remove the control-rate parameter from all ```startMozzi()``` calls in examples. It used to be required, there, if ```MOZZI_CONTROL_RATE``` was customized, but it no longer is, when moving the define up in the sketch. So also in that respect, ```MOZZI_CONTROL_RATE``` behave more consistent with other config defines, now. While the parameter still exists in ```startMozzi()``` its usage is no longer encouraged.
- Fixed a compiler warning, and moved around some info.


tfry-git:
As mentioned, this is just a collection of small things, but some of these changes are quite "noisy", and so I don't think it make sense to stack up much more in this PR. Rather, I'd like to merge this bunch into devel/Mozzi2, then start fresh with other tidbits on the TODO list. Before I do, could you take a look?

I listed the changes in the summary.

AudioOutput.h
```diff
@@ -52,72 +52,35 @@
  *  than 16 bits). */
 #define AudioOutputStorage_t int
 
-#define SCALE_AUDIO(x,bits) (bits > MOZZI_AUDIO_BITS ? (x) >> (bits - MOZZI_AUDIO_BITS) : (x) << (MOZZI_AUDIO_BITS - bits))
-#define SCALE_AUDIO_NEAR(x,bits) (bits > MOZZI_AUDIO_BITS_OPTIMISTIC ? (x) >> (bits - MOZZI_AUDIO_BITS_OPTIMISTIC) : (x) << (MOZZI_AUDIO_BITS_OPTIMISTIC - bits))
-#define CLIP_AUDIO(x) constrain((x), (-(AudioOutputStorage_t) MOZZI_AUDIO_BIAS), (AudioOutputStorage_t) MOZZI_AUDIO_BIAS-1)
+template<typename T> constexpr AudioOutputStorage_t SCALE_AUDIO(T x, byte bits) { return (bits > MOZZI_AUDIO_BITS ? (x) >> (bits - MOZZI_AUDIO_BITS) : (x) << (MOZZI_AUDIO_BITS - bits)); }
```
tomcombriat, line 55: I do not know why, but I had this wrong idea that ternary was not allowed in `constexpr`, learning new things everyday (which might be useful for FixMath also…

AudioOutput.h
```diff
@@ -52,72 +52,35 @@
...
-#define AudioOutput_t AudioOutputStorage_t
-/** Representation of an single audio output sample/frame. This #define maps to either MonoOutput or StereoOutput, depending on what is configured
+/** Representation of an single audio output sample/frame. This typedef maps to either MonoOutput or StereoOutput, depending on what is configured
  *  in MOZZI_AUDIO_CHANNELS. Since the two are source compatible to a large degree, it often isn't even necessary to test, which it is, in your code. E.g.
  *  both have functions l() and r(), to return "two" audio channels (which will be the same in case of mono).
  *
  *  You will not usually use or encounter this definition, unless using @ref external_audio output mode.
  */
-#define AudioOutput MonoOutput
+typedef MonoOutput AudioOutput;
```
tomcombriat, line 71: MonoOutput is still an `int` internally right?

examples/03.Sensors/Piezo_Frequency/Piezo_Frequency.ino
```diff
@@ -26,24 +26,23 @@
    Tim Barrass 2013, CC by-nc-sa.
 */
 
+// increase the rate of updateControl from the default of 64, to catch the piezo's rapid transients
+#define MOZZI_CONTROL_RATE 150
```
tomcombriat, line 30: I had the idea that `CONTROL_RATE` had to be a power of two… I do not see at the top of my head why right now but we might change that to 128 or 256?

tomcombriat (approved):
Looks good to me! I have to say that I skimmed the examples a bit…

AudioOutput.h
tfry-git: No, but yes ;-). It's a class, but with no virtual functions, and an int as only data member, which means it is stored as an int. It also has implicit conversion to, and from int, so flash/RAM usage stays the same, and everything that used to be possible still is,.

*Except* when MOZZI_COMPATIBILITY_LEVEL is set to greater than 1.1 (which, by default, is the case, when including Mozzi.h, but not when including MozziGuts.h), defining updateAudio() as returning int is no longer accepted.

examples/03.Sensors/Piezo_Frequency/Piezo_Frequency.ino
tfry-git: I hesitated a second on that one, too. Strictly speaking, I guess the requirement to use only powers of two only really applies where the ratio of audio-/control-rate comes into play, somewhere. But yes, I guess you are right, and we simply should encourage something like this in the examples. Will adjust.

PR #233, opened 2024-02-01
Implementing auto Range for FixMath

tomcombriat:
This implements what was discussed in #212 : promoting `NI` of `FixMath` only when needed. 

The approach I am landing now is to keep track of the `RANGE` actually used by the number, initialized to the maximum holdable value by default.

This will probably make heavy use of Macro, because tests will be involved (not available in C++11 constexpr), but that should be a big deal as they are just template parameters which are not used for any calculation.

This draft PR is just to keep an eye on compilations and memory usage automatically.

### DIFF
--- a/FixMath.h
+++ b/FixMath.h
@@ -17,43 +17,68 @@
     signed (SFixMath) or unsigned (UFixMath).
 
     A fixed point number has its range defined by the number of bits encoding the integer part (NI 
-    in the following) and its precision by the number of bits encoding the fractional part (NF). For UFixMath types, the integral part can hold values in [0,2^NI-1], for SFixMath types, the integral part can hold values in [-2^NI,2^NI-1].
+    in the following) and its precision by the number of bits encoding the fractional part (NF). For UFixMath types, the integral part can hold values in [0,2^NI-1], for SFixMath types, the integral part can hold values in [-2^NI,2^NI-1]. The number of bits encoding the fractional can be considered as the precision of the number: given NF, the number of possible values in the  [0,1[ range will 2^NF. Hence, given NF, the resolution will be 1/(2^NF).
+
+    Under the hood, these types will keep track of the maximum possible value they might hold (this is the RANGE template parameter), and, if only *SAFE* operations (see below) are used, will automatically adjust there NI and NF to accomodate the result of a operation. It will also try not to promote there internal type when possible, assuming that you use the complete range of a given type.
+
+    The operations possible with these types can be divided into two categories:
+    - the operations between FixMath types are all safe (aka won't overflow) and are the only one included by default
+    - the operations between a FixMath and a native C type (int, float) are NOT safe and are not included by default. In order to activate them, you need to `#define FIXMATHUNSAFE` before including FixMath.h.
+
 
     Like standard C(++) types, the fixed point numbers defined here are following some rules:
     - any fixed type can be converted to another *as long as the value can be represented in the destination type*. Casting to a bigger type in term of NI and NF is safe, but reducing NI can lead to an overflow if the new type cannot hold the integer value and reducing NF leads to a loss of precision.
+    - Fixed types can be constructed from and converted to standard C types.
     - all operations between fixed point number is safe (it won't overflow) and preserve the precision. In particular:
-    - only addition, subtraction and multiplication are implemented
+    - only addition, subtraction and multiplication are implemented (this is a design choice, see below)
     - any operation between a signed and an unsigned leads to a signed number
     - resulting numbers will be casted to a type big enough to store the expected values. It follows that it is worth starting with types that are as small as possible to hold the initial value.
-    - all operations between a fixed point number and a native type (int, float, uint) are *not* safe. If the resulting value cannot be represented in the fixed point type it will overflow. Only addition, subtraction, multiplication and right/left shift are implemented.
-    - safe right/left shifts, which return the correct value in the correct type are implemented as .sR<shift>() and .sL<shift>() respectively, shift being the shifting amount. These shifts are basically on
+    - all operations between a fixed point number and a native type (int, float, uint) are *not* safe. If the resulting value cannot be represented in the fixed point type it will overflow. Only addition, subtraction, multiplication and right/left shift are implemented. These are only accessible activating the `FIXMATHUNSAFE` set.
+    - safe right/left shifts, which return the correct value in the correct type are implemented as .sR<shift>() and .sL<shift>() respectively, shift being the shifting amount.
 
     More specifically on the returned types of the operations between fixed point math types:
     - Additions:
-    - UFixMath<NI,NF> + UFixMath<_NI,_NF> returns UFixMath<MAX(NI,_NI)+1,MAX(NF,_NF)>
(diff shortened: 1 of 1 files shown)


tomcombriat:
Okay, this seems to work! Only addition for `UFixMath` is done for now. It does not even seem to be so much trouble, so far. (Well, the compiler seems to produce quite some big lines to figure the `RANGE` but well, that's its job I guess!).

This version add the `RANGE` when performing an addition, and if the range exceed the biggest container, it promotes to the next one (but keep track of what is used in that one).

With that version:
```
UFixMath<8,0> a, b, c, d;
auto e = a+b+c+d;
```
correctly makes `e` a `UFixMath<10,0>` and not a `UFixMath<11,0>`.

Be patient for the rest, the lines do not turn that difficult, but I am trying to get my head not messing up the algebra too bad…

tomcombriat:
Ok, slowly getting there… Not as straightforward as I initially thought (you can have a look at the MACRO I had to use…).
Nearly everything is done, except the multiplication and opposite for `SFix`.

One of the small difficulties funnily (maybe for me only ;) ) comes from the fact that we are using two's complement for negative numbers. It follows that the attainable range is not exactly the same for unsigned (255 in 8bits) and signed (256 in 8 coding bits). But I took that into account, so for instance, the library knows that, in the following example, `c` can only go to 127*2=254 and not 255.
```
UFixMath<8,0> a, b;
auto c = a+b; 
 ```
 The gain is not very impressive on one addition but that can allow to squeeze one more small number in some cases, after a few addition.

---

Will try to finish, clean up and have a look with a fresh eye soon (a quick glance from someone else might be helpful there too), I discovered some errors I did in the previous implementation already (especially in the "smart" shift, which should be careful if `NI` or `NF` goes to 0 because of the shift)…

---

PS: do not understand why the reports do not work, I started from the `devel/FixMath2` branch which was up-to-date with Github actions…

**Comment by tfry-git** [comment]
I looked at your changes, and could not find anything wrong, but it's quite complex stuff, indeed.

Regarding the NEEDEDNIEXTRA macro, I wonder, whether NI is actually needed as a parameter, here? Isn't it rather used as "NIFORRANGE", essentially?

At any rate, I would highly welcome some automated tests. One way to implement those might be as follows (directly in FixMath.h):
```
namespace MozziPrivate {
void testUFix() {  // Just a dummy. This function will never be called.
  const UFixMath<7, 0> a = 66;    // as long as these are declared const, we can use them in static_assert()s
  const UFixMath<0, 1> b = .5;

  static_assert((a*b).SR(1).asRaw() == 33);
  static_assert((a+b).getNI() == 7);
  static_assert((a+a+a).getNI() == 8);
  // many more tests to be added, here, over time
  // both to document expected behavior, and to make sure future changes don't break it
}
}
```

(While typing the above, I was looking for an ```asInt()```-function - perhaps we decided against that in order to avoid confusion with ```asRaw()```, but how about an ```integerPart()``` getter, perhaps.)

--

No idea, why reports don't work. Perhaps comparison across forks is an issue, here?

--

Regarding the plans, hiding away the unsafe functions, until explicitly requested, sounds good to me.

Yes, we'll need casting within the S/UFix types. I still think it might be possible to come up with a set of member functions that makes some such casts easier to read - if designed, properly. (You remember my half-baked idea on clip() and trim(); floor() (shift out by NF) is another one that I imagine could be useful).

As for correct initial choice of types, it would be cool if we could somehow coerce the compiler to apply a proper RANGE, when initializing from a constant. Not sure, if that can be done.

**Comment by tfry-git** [comment]
Except of course, my static_asserts don't seem to compile. I'll look into how this is (hopefully) possible.

**Comment by tfry-git** [comment]
Ignore my comment regarding NEEDEDNIEXTRA: I somehow took NI+NF as a given, and then concluded, NI was redundant...

---

> Too bad macros cannot be recursive

Templates can ;-)

```
template<int64_t value, int8_t bits=0> class BitCounter {
public:
    static constexpr char countbits() {
       return((1LL << bits) > value ? bits : (BitCounter<value, bits+1>::countbits()));
   }
};
// partial specialization to stop counting at 64
template<int64_t value> class BitCounter<value, 64> {
public:
    static constexpr char countbits() { return 64; }
};

static_assert(BitCounter<127>::countbits() == 7);
```

---

Regarding my static_asserts: These can be made to work by making a lot of stuff constexpr.
 - For a lot of the functions that may actually be useful in its own right.
 - The c'tor would also to be constexpr, which (in c++11) means no function body allowed at all: 
    ```
    template<typename T>
    constexpr UFixMath(T value, bool as_raw) : internal_value(as_raw ? internal_type(value) << NF : value) {}
    ```
- Where it gets a bit nasty is the + and * operators, however. Again, C++11 strictly insists on "only a return statement",
  which is quite possible, but at the cost of readability. 
- Finally, the compiler also seems to need a few extra "constexpr" thrown in in the test code itself (e.g. ```  const auto constexpr a = UFixMath<7, 0>::fromRaw(66)```, for whatever reason.

So, all in all, this is again more complex than expected (even if I tend to think it may be worth it in the long run), and I suggest to ignore it for the moment.

tomcombriat:
Alright, I think it looks quite close to a final first step.

The unsafe functions are deactivated by default, they can be activated by `#define FIXMATHUNSAFE` before including `FixMath.h` (I am quite sure you will come with a cleaner solution).

Remains: 
 
- [ ] Documentation
- [ ] Adding `toSFix` for `UFix` and the opposite: safe casting to the other type which also keeps the `RANGE` (which I think is not possible with the standard cast defined in constructors).

tomcombriat:
> Templates can ;-)

Good catch! I add that to my ever-growing list!

---

> The c'tor would also to be constexpr, 

Is that possible even if the arguments taken by the constructor are *not* `const` (like it would happen when say, casting from any changeable value like `mozziAnalogRead()`)? Or two versions would be needed?

Also, on readability, the places mixed operators are defined follows somehow a logic (in UFix except if the code is more similar to SFix) but I could agree it is not straightforward, especially as their respective opposite are defined as free functions at the end to avoid code duplication (they can be defined by calling the functions with reversed operators).
All mixed comparisons are together at the end somehow, but I found it quite clean there because they are quite simple, compared to the operators…

tfry-git:
> > Templates can ;-)
> 
> Good catch! I add that to my ever-growing list!

Note that I'm not sure, how this would be applied towards detection of NI, however, esp. in light of the following:

> > The c'tor would also to be constexpr,
> 
> Is that possible even if the arguments taken by the constructor are _not_ `const` (like it would happen when say, casting from any changeable value like `mozziAnalogRead()`)? Or two versions would be needed?

```constexpr``` is just a promise that the statement can be evaluated at compile-time, *if* all relevant parameters are known at compile time. It may be helpful to know that there is also a - different - ```consteval``` (C++20 or so, far out of reach), which enforces that a function may *only* be evaluated at compile-time.

So one constructor is still enough.

Also, the BitCounter-template could also be made to work on a variable input, but it would produce terribly inefficient code in that case.

tomcombriat:
(Replaced all the macro with `constexpr`, now that I know ternary are allowed there… I hope to get the extra point ;) ).

Cleanup here and documentation in #212 (once merged) still needed.

PR #236, opened 2024-02-18
Specializations of Mozzi classes for U/SFix

tomcombriat:
Still work in progress. So far only `Line` and `Smooth` are done.

I am a bit sad that there is so much code duplication between `Line<UFix>` and `Line<SFix>` but up to now, I did not figure a way to avoid that. Maybe a virtual mother class `FixType`, which both `UFix` and `SFix` would inherit from would be the solution? (I read here and there that the performances of such dynamic polymorphism are not as good, except by doing Curiously recurring template pattern maybe?)


tfry-git:
> I am a bit sad that there is so much code duplication between `Line<UFix>` and `Line<SFix>` but up to now, I did not figure a way to avoid that.

I suppose you tried some stuff (while I did not). But first question: Is a full specialization needed at all, won't the default implementation cover most of this?

(Side note: Those ```volatile``` specifiers should be long since obsolete, and should be removed). 

Runtime polymorphism would surely impose a performance hit, indeed (going quite counter to the whole idea of UFix/SFix). Toying with the idea, however, maybe there could be a common template:

```
template<int8_t NI, int8_t NF, bool SIGNED, uint64_t RANGE> class GFix {
// full implementation for UFix and SFix inside, possibly, however in the form a common declaration, and then
// separate (partial) specializations for the member functions
};
```

And after than, UFix and SFix would simply be given as template aliases:

```
template<int8_t NI, int8_t NF, uint64_t RANGE>
using SFix = GFix<NI, NF, true, RANGE>;

template<int8_t NI, int8_t NF, uint64_t RANGE>
using UFix = GFix<NI, NF, false, RANGE>;
```

I admit, that looks a bit daunting...

tomcombriat:
Thanks for your comment!

> I suppose you tried some stuff (while I did not). But first question: Is a full specialization needed at all, won't the default implementation cover most of this?

Indeed, only some functions might need to be specialized (I copied the way the specialization is done), but they are the interesting/complex ones (but implementation is the same for U/SFix).

To be fair, this only concerns `Line` and `Smooth`, which are already done, maybe I am overthinking it… (And I am not sure I am ready for `template<int8_t NI, int8_t NF, bool SIGNED, uint64_t RANGE> class GFix` at this time ;) ). 

--- 

> (Side note: Those volatile specifiers should be long since obsolete, and should be removed).

Will do. I was actually wondering about them.

tomcombriat:
Small note while I am here: I am now in the process to convert the examples to `FixMath`. It is not always as easy as it seems as some of the original examples build heavily on shifts to fit things correctly into old fixmath types. All this logic needs to be taken apart.

Note that I sometimes keep the original ones as well, for now for historical reasons, we can decide later if we want to replace everything or keep some of them.

Note also that, in the `Difference_Tone` example, `Q16n16 freq1 = Q8n0_to_Q16n16(440);` leads to a frequency of 184 Hz actually as 440 cannot be hold in 8bits. That explains this nice bass tone on this example.

examples/06.Synthesis/Detuned_Beats_Wash/Detuned_Beats_Wash.ino
```diff
@@ -110,47 +116,72 @@ void updateControl(){
...
+*/
+  auto asig = 
+  toSFraction(aCos1.next()) + toSFraction(aCos1b.next()) +
+  toSFraction(aCos2.next()) + toSFraction(aCos2b.next()) +
+  toSFraction(aCos3.next()) + toSFraction(aCos3b.next()) +
+  toSFraction(aCos4.next()) + toSFraction(aCos4b.next()) +
+  toSFraction(aCos5.next()) + toSFraction(aCos5b.next()) +
+  toSFraction(aCos6.next()) + toSFraction(aCos6b.next()); /* +
+toSFraction(aCos7.next()) + toSFraction(aCos7b.next()) +*/
+  return MonoOutput::fromSFix(asig);
```
tomcombriat, line 185: @tfry-git Here comes the automatic output scaling :)

tomcombriat:
There seems to be a slight cost of using `MonoOutput::fromSFix` (20bits on the `Detuned_Beats_Wash` example… But I do not manage to find where… It apparently comes from `toSFraction()`…

tomcombriat:
(I do not understand why the actions do not find some of the examples…)

AudioOutput.h
```diff
@@ -46,6 +46,7 @@
 
 #ifndef AUDIOOUTPUT_H
 #define AUDIOOUTPUT_H
+#include <FixMath.h>
```
tfry-git, line 49: On a somewhat general note, I wonder, whether we can still avoid a hard dependency on the - now external - FixMath. Sure, installing it via the library manager should be a breeze, but at least when installing from git, that may be a small bonus.

The trick would be to never have a *full* specialization. Due to the RANGE parameter, I think that requirement is actually easy to meet. Pseudocode:

```
//#include <FixMath.h> -> removed
template<int8_z, int8_t, uint64_t> class SFix;  // forward declaration, only

// This partial specialization is only compiled when actually used
// therefore it's ok that SFix is an incomplete type, so far.
template<int8_t NF, uint64_t RANGE> void someFunction(SFix<16,NF,RANGE> x) { [...] };
```

Oscil.h
```diff
@@ -157,9 +157,9 @@ class Oscil
 	each direction. This fixed point math number is interpreted as a SFix<15,16> internally.
```
tfry-git, line 157: Optimization question (maybe for later): Aren't those 15 NI bits fully ineffective, 16 fractional bits alone would fully suffice? (Sign might be a problem, I imagine)

examples/05.Control_Filters/Line_vs_Smooth/Line_vs_Smooth.ino
```diff
@@ -60,29 +60,30 @@ Smooth <unsigned int> aSmooth(smoothness); // to smooth frequency for aSin1
...
+  UFix<16,16> freq0 = mozziAnalogRead(0); // 0 to 1023, with an additionnal 16bits of precision (which will be used in the interpolation.)
   freq1 = (unsigned int) mozziAnalogRead(1); // 0 to 1023
-   aInterpolate.set(freq0, AUDIO_STEPS_PER_CONTROL);
 }
 
 
 AudioOutput updateAudio(){
-  Q16n16 interpolatedFreq = aInterpolate.next(); // get the next linear interpolated freq
-  aSin0.setFreq_Q16n16(interpolatedFreq);
+  UFix<16,16> interpolatedFreq = aInterpolate.next(); // get the next linear interpolated freq
```
tfry-git, line 78: Style: Maybe use auto, here?

examples/05.Control_Filters/Line_vs_Smooth/Line_vs_Smooth.ino
```diff
@@ -60,29 +60,30 @@ Smooth <unsigned int> aSmooth(smoothness); // to smooth frequency for aSin1
...
 void setup(){
-  aSin0.setFreq(660.f);
-  aSin1.setFreq(220.f);
+  aSin0.setFreq(660);
+  aSin1.setFreq(220);
   startMozzi();
 }
 
 
 volatile unsigned int freq1;  // global so it can be used in updateAudio, volatile to stop it getting changed while being used
```
tfry-git, line 69: volatile can be removed, again

examples/05.Control_Filters/Line_vs_Smooth/Line_vs_Smooth.ino
```diff
@@ -60,29 +60,30 @@ Smooth <unsigned int> aSmooth(smoothness); // to smooth frequency for aSin1
...
 
 void updateControl(){
-  Q16n16 freq0 = Q16n0_to_Q16n16(mozziAnalogRead(0)); // 0 to 1023, scaled up to Q16n16 format
+  UFix<16,16> freq0 = mozziAnalogRead(0); // 0 to 1023, with an additionnal 16bits of precision (which will be used in the interpolation.)
   freq1 = (unsigned int) mozziAnalogRead(1); // 0 to 1023
-   aInterpolate.set(freq0, AUDIO_STEPS_PER_CONTROL);
```
tfry-git, line 74: Isn't this line still needed?

examples/06.Synthesis/Detuned_Beats_Wash/Detuned_Beats_Wash.ino
```diff
@@ -26,74 +26,80 @@
...
 
 // harmonics
-Oscil<COS8192_NUM_CELLS, MOZZI_AUDIO_RATE> aCos1(COS8192_DATA);
-Oscil<COS8192_NUM_CELLS, MOZZI_AUDIO_RATE> aCos2(COS8192_DATA);
-Oscil<COS8192_NUM_CELLS, MOZZI_AUDIO_RATE> aCos3(COS8192_DATA);
-Oscil<COS8192_NUM_CELLS, MOZZI_AUDIO_RATE> aCos4(COS8192_DATA);
-Oscil<COS8192_NUM_CELLS, MOZZI_AUDIO_RATE> aCos5(COS8192_DATA);
-Oscil<COS8192_NUM_CELLS, MOZZI_AUDIO_RATE> aCos6(COS8192_DATA);
-//Oscil<COS8192_NUM_CELLS, MOZZI_AUDIO_RATE> aCos7(COS8192_DATA); // used to work smoothly in Arduino 1.05
+Oscil<COS8192_NUM_CELLS, AUDIO_RATE> aCos1(COS8192_DATA);
```
tfry-git, line 40: That is still allowed, but why did you revert from MOZZI_AUDIO_RATE, here?

mozzi_midi.h
```diff
@@ -113,21 +113,34 @@ inline Q16n16  Q16n16_mtof(Q16n16 midival_fractional)
...
 {
   return UFix<16,16>::fromRaw(Q16n16_mtof(midival.asRaw()));
-}
+};
 
-template<int8_t NI, int8_t NF>
-  inline UFix<16,16> mtof(UFix<NI,NF> midival)
+/** @ingroup midi
+Converts midi note number with speed and accuracy from any UFix.
+Uses Q16n16_mtof internally.
```
tfry-git, line 128: Optimization: A specialization for the case of plain midi notes (midival with NF=0) may avoid unexpeced inefficiency in many cases.

tfry-git (commented):
A quick review so as not to keep you waiting too much longer. I did not look super-closely at the examples, and importantly, I did not test even a single one...

AudioOutput.h
tfry-git: And to clarify, FixMath.h would then have to be included, manually, in sketches that actually use it.

tomcombriat: That's a good idea, I did not think of it. My only concern would be that the dependency is less clear (ie, the error message would be something like: `SFix is declared but never defined` and not `File FixMath.h not found`). But at the same time, the error message for the examples would be clear. Will try to see how that goes!

Oscil.h
tomcombriat: Yes, I thought of that too. Some examples (for instance FMSynth) do wander further than pure fractional, but as it rolls back, only fractional should be enough (and further is taking care of  smoothly by the overflow). Indeed the sign is a problem, but maybe an extra function taking SFix<0,15>, so sacrificing one bit of precision, could be interesting. Might not be straightforward to make the compiler choose between the two though. 

I would suggest taking that a bit further down the pipeline as this is more optimisation of legacy code that directly related to FixMath.

examples/05.Control_Filters/Line_vs_Smooth/Line_vs_Smooth.ino
tomcombriat: Yes indeed.

examples/06.Synthesis/Detuned_Beats_Wash/Detuned_Beats_Wash.ino
tomcombriat: Old habits are sometimes hard to get rid off… I changed it back!

mozzi_midi.h
tomcombriat: I'll look it up. I might go through an optimization with Q8n0, that would be call upon (in the same spirit than what is done here with Q16n16).

tomcombriat:
Thanks for the comments! I think the optimizations can also be taken care of further down the line (I though the same when working on mtof) but I will start thinking on it. If I come up with a nice solution (might be very easy) before we want to merge I will it here.

mozzi_midi.h
tomcombriat: Done

AudioOutput.h
tomcombriat: I might be overlooking something, but I am not sure this is always simple. Let's take this line from Oscil.h:

```
        template <int8_t NI, int8_t NF, uint64_t RANGE>
        inline
        void setFreq(UFix<NI,NF,RANGE> frequency)
        {
        setFreq_Q16n16(UFix<16,16>(frequency).asRaw());
        }
```

with the forward declaration `setFreq(UFix<NI,NF,RANGE> frequency)` is correctly resolved by the compiler but `UFix<16,16>(frequency)` is not as it does not have the correct number of arguments. Adding another argument, say `_RANGE`, is not straightforward because we would need to add in Mozzi the macros to compute the default range here (hence negating a bit the advantage of having it with a default value in FixMath.

Did I overlooked something?

tfry-git: > Did I overlooked something?

Only the fact that I tend to comment before testing... It may or may not be possible to weasle our way out of this particular problem using something like

```
setFreq_Q16n16((int32_t) frequency.sL(16-NF).asRaw());
```

But I'm not sure, where the next problem of this type pops up, and I'll admit, I'm not totally convinced, myself.

Oscil.h
tfry-git: Agreed. Maybe collection things like this somewhere would help keeping track, though. Perhaps in an issue, or by marking them as "TODO: optimize" in the code.

tomcombriat:
Anything that should be done here before merging? I am starting to be afraid that this might diverges quite a bit (especially with #239 and #240)

tfry-git:
> Anything that should be done here before merging? I am starting to be afraid that this might diverges quite a bit (especially with #239 and #240)

Not that I am aware of. You are right in that we should try to put the pieces together, now.

tomcombriat:
Top! I won't be able to work on it in the next days but will try to get it up to date when I come back for merging!

PR #238, opened 2024-02-25
Removed FixMath from Mozzi's sources

tomcombriat:
This is to check if the automatic tests are okay with an external dependency… 
~The tests seem to fail but I wonder if this is not because the `library.properties` it uses is the "old" one, without the dependency…~

~Tests are happy, but the examples are not actually compiled…~


tomcombriat:
Merging this.

PR #239, opened 2024-03-24
Scaling for analog reads

tfry-git:
As discussed, this adds scaling to analog reads.


MozziGuts.h
```diff
@@ -136,7 +136,15 @@ are called.
 */
 void audioHook();
 
+/** @ingroup analog
+
+See getAudioInput(). The template parameter specifies the desired value range in bits. */
+template<byte RES> int16_t getAudioInput();
```
tomcombriat, line 142: I might be missing something, but I would have expected everything to be `uint16_t` and not `int16_t`: analogReads are always positive.

MozziGuts.h
```diff
@@ -150,10 +158,19 @@ and
...
+
 @return audio data from the input buffer
 */
-#if !MOZZI_IS(MOZZI_AUDIO_INPUT, MOZZI_AUDIO_INPUT_NONE)
-int getAudioInput();
+#if defined(FOR_DOXYGEN_ONLY) || (!MOZZI_IS(MOZZI_AUDIO_INPUT, MOZZI_AUDIO_INPUT_NONE))
+#if defined(FOR_DOXYGEN_ONLY) || defined(MOZZI_ANALOG_READ_RESOLUTION)
+inline int16_t getAudioInput() { return getAudioInput<MOZZI_ANALOG_READ_RESOLUTION>(); };
+#else
+MOZZI_DEPRECATED("2.0", "This use of getAudioInput() is not portable") inline int16_t getAudioInput() { return getAudioInput<MOZZI__INTERNAL_ANALOG_READ_RESOLUTION>(); };
```
tomcombriat, line 172: Maybe adding to the warning what is the "good practice" could be useful?

internal/MozziGuts.hpp
```diff
@@ -27,6 +27,11 @@ static void advanceADCStep();                       // to be provided by platfor
 static void startSecondADCReadOnCurrentChannel();   // to be provided by platform implementation
 static uint8_t adc_count = 0;                       // needed below
 #endif
+
+// TODO: make this helper public?
+template<byte BITS_IN, byte BITS_OUT, typename T> constexpr T smartShift(T value) {
+    return (BITS_IN > BITS_OUT) ? value >> (BITS_IN - BITS_OUT) : (BITS_IN < BITS_OUT) ? value << (BITS_OUT - BITS_IN) : value;
```
tomcombriat, line 33: This one rings a bell ;)

internal/MozziGuts.hpp
```diff
@@ -126,15 +131,15 @@ __attribute__((noinline)) void adcStartReadCycle() {
...
+int16_t mozziAnalogRead(uint8_t pin) {
 	pin = adcPinToChannelNum(pin); // allow for channel or pin numbers; on most platforms other than AVR this has no effect. See note on pins/channels
 	adc_channels_to_read.push(pin);
 	return analog_readings[channelNumToIndex(pin)];
 }
 
 #if !MOZZI_IS(MOZZI_AUDIO_INPUT, MOZZI_AUDIO_INPUT_NONE)
 static AudioOutputStorage_t audio_input; // holds the latest audio from input_buffer
-AudioOutputStorage_t getAudioInput() { return audio_input; }
+int16_t getAudioInput() { return audio_input; }
```
tomcombriat, line 142: Idem `uint16_t`?

internal/config_checks_avr.h
```diff
@@ -133,6 +133,8 @@
 #define MOZZI_AUDIO_BITS (2*MOZZI_AUDIO_BITS_PER_CHANNEL)
 #endif
 
+#define MOZZI__INTERNAL_ANALOG_READ_RESOLUTION 10
```
tomcombriat, line 136: One thing is that, on some platform, the ADC resolution is editable (`analogReadResolution()`) maybe this define should be editable like other Mozzi's config at the top of the user sketch. In that case, a user can let mozzi knows of the actual ADC resolution? It would also solves partly the problem of mbed, where all platforms might not have the same resolution?

tomcombriat (commented):
Looks good to me! I let a few minor comments here and there that you can probably ignore. We should not forget to adapt the few examples.

MozziGuts.h
tfry-git: Good point. This used to be int, and I figured I'd define that more precisely, while at it. Noteably, standard analogRead() returns int. However using int16_t (or int on AVR) would get us into trouble when requesting 16 bits. I  changed this to uint16_t

MozziGuts.h
tfry-git: Sort of hard to put that into a few words. I added a pointer to the documentation, though.

internal/MozziGuts.hpp
tfry-git: Done.

internal/config_checks_avr.h
tfry-git: I think the way to implement that will be:
- user defines MOZZI_ANALOG_READ_RESOLUTION
- implementation can see this, and may decide to set the internal resolution, accordingly. Approximately:

```
#ifdef MOZZI_ANALOG_READ_RESOLUTION
#define MOZZI__INTERNAL_ANALOG_READ_RESOLUTION MOZZI_ANALOG_READ_RESOLUTION
#define MOZZI__IMPL_SET_ANALOG_READ_RESOLUTION
#endif

[...]

//inside setupMozziADC
#ifdef MOZZI__IMPL_SET_ANALOG_READ_RESOLUTION
analogReadResolution(MOZZI_ANALOG_READ_RESOLUTION);
#endif
```

The advantages are that only one define is needed, implementations are free to set this internally, or rely solely on shifting, and users won't need to worry about which it is.

tomcombriat: Oh, that is even one step further than what I though (I was thinking that the user sets `analogReadResolution` in the user code and let Mozzi know about it with the define). I guess the `analogReadResolution` should be portable so probably even better to put it there!

MozziGuts.h
tomcombriat: I think this one escaped :P

internal/MozziGuts.hpp
tomcombriat: I think l. 134 and l.142 should also be unsigned?

internal/MozziGuts.hpp
```diff
@@ -289,9 +294,9 @@ uint32_t MozziRandPrivate::z=521288629;
 unsigned long audioTicks() { return MozziPrivate::audioTicks(); };
```
tomcombriat, line 294: Sorry, but looking at this file (I am still lost in the new organization, too many guts), and this is not related, but why would `unsigned long mozziMicros() { return MozziPrivate::mozziMicros(); };` conflict with `mozziMicros()` but not the other ones?

tfry-git: Oh, good catch. I must have commented that line by accident. Fixed (and also clarified the explanation at bit).

internal/config_checks_avr.h
tfry-git: Note that we may or may not take readings at the ```analogReadResolution()```, as we may or may not utilize the same mechanisms as ```analogRead()```. The pico, for instance does not actually seem to support requesting anything other than 12 bits from the hardware, and ```analogRead()``` merely shifts that at runtime. As we're not using that, we always need to shift from 12 bits, irrespective of analogReadResolution(). I _believe_ on AVR, we're similarly working at a fixed resolution, internally.

STM32(duino) and Mbed (via Arduino_AdvancedAnalog) do seem to support hardware reads at different resolution (again independent from ```analogReadResolution()```), although not all bit values are possible (no 13 bits, for instance), and I'm a bit reluctant about touching the already complex code, here.

So what I've outlined above is a really just theory for the time being. I'll add a note to that effect in the template files, though.

internal/MozziGuts.hpp
tfry-git: Yes, indeed. And of course line 141. I hope we've covered everything, now...

tomcombriat (approved):

Issue #248, opened 2024-04-07
I2S dacs config in Mozzi2


poetaster:
I had tested several with mozzi 1.x and had success with the PT8211 and the PCM5102 with mozzi. I hadn't tried the MAX9898357 yet, but it worked with pschatzmann's audio-tools.

Now, I've switched to Mozzi2 and I can't seem to get I2S working on the rp2040, pico. But, I'm probably just doing it wrong. The configuration has changed a lot, so I just tried to follow the PT8211 examble.  I checked the Datasheet on the 5102 first: https://www.ti.com/lit/ds/symlink/pcm5102.pdf It states:  the audio input interface format is left justified and Audio Data format, MSB First, 2’s Complement. I proceeded with:

```// before including Mozzi.h, configure external audio output mode:
#include "MozziConfigValues.h"  // for named option values
#define MOZZI_AUDIO_MODE MOZZI_OUTPUT_I2S_DAC
#define MOZZI_AUDIO_CHANNELS MOZZI_STEREO
#define MOZZI_AUDIO_MODE MOZZI_OUTPUT_EXTERNAL_TIMED
// all of these are the defaults on RP2040 outputting on I2S, anyway, thus commented
#define MOZZI_AUDIO_BITS 16
#define MOZZI_I2S_PIN_BCK 20
#define MOZZI_I2S_PIN_WS (MOZZI_I2S_PIN_BCK+1) // CANNOT BE CHANGED, HAS TO BE NEXT TO pBCLK, i.e. default is 21
#define MOZZI_I2S_PIN_DATA 22
#define CONTROL_RATE 256 // Hz, powers of 2 are most reliable
#include <Mozzi.h>
````
My hardware designs for the rp2040 are all setup so that 20,21, 22 are free to be assigned this was (can alternately be used for sda/scl video and 22 for pwm). 

This does not work. What I missing.  Do I need to specify some combination of:
```#define MOZZI_I2S_FORMAT MOZZI_I2S_FORMAT_LSBJ // PT8211 is on LSBJ format
```
or
```
MOZZI_I2S_FORMAT_PLAIN

``` ??

I had tried LSBJ just for the hell of it, but wasn't certain.

tfry-git:
> #define MOZZI_AUDIO_MODE MOZZI_OUTPUT_I2S_DAC
> #define MOZZI_AUDIO_MODE MOZZI_OUTPUT_EXTERNAL_TIMED

[...]

PR #249, opened 2024-04-09
Example sketch for stereo I2S DACs

poetaster:
Since it took a bit of effort, I thought I'd supply a concrete example sketch.


poetaster:
Oh, dear. Sorry for all the noise. I thought I'd do a good deed and in the process pulled devel/Mozzi2 ... now I get  fatal error: FixMath.h: No such file or directory  Any tip how to revert ?

tomcombriat:
Thanks for the PR! Cool to know that other DACs are working!
As per the example, maybe it would make more sense to add it as config example (in the spirit of what is in config/)?

poetaster:
> Hi, This is normal since #236, Mozzi is now using another library for fixed point arithmetics: FixMath. So far it is a hard dependency. You just need to install it, either in the Arduino library manager, or via [github](https://github.com/tomcombriat/FixMath).

Ah, thanks for clarifying. !

poetaster:
> Thanks for the PR! Cool to know that other DACs are working! As per the example, maybe it would make more sense to add it as config example (in the spirit of what is in config/)?
> 
> Does this close #248 ?

Yes, sure thing. I had noted a bit of noise which I documented with an upload, but I'll try to investigate that with different hardware setups and let you know. I 'll go install that library :)

tomcombriat:
(Sorry for the fuss, was a bit fast to delete the branch this was pointing to… Will try to test soon!)

PR #253, opened 2024-05-05
Non inverting ADSR

tomcombriat:
For some values of times, when using an `ADSR` updated at `MOZZI_CONTROL_RATE`, the approximation made in `Line` to compute the number of steps sometimes results in an overflow of the envelope. For instance, a long decay phase aiming for a 0 value, briefly jump to 255 (the Line does not overflow, but the conversion to `unsigned` does). This appears only for some time values, but when it does it is quite obvious (see image).

This solves this by increasing the planned number of control steps for the `Line` to reach the value (it decreases the slope of the transition slightly), to make sure that we reach an update before this phenomenon happens.

It changes very slightly the length of the envelopes (by 1/`MOZZI_CONTROL_RATE`) but this is very small compared to the error in the timing of the envelope (where a shift is used to divide by 1000).

## Testing sketch:

```
#define MOZZI_CONTROL_RATE 64
#include <Mozzi.h>
#include <Oscil.h>
#include <EventDelay.h>
#include <ADSR.h>
#include <tables/sin8192_int8.h>

Oscil<8192, MOZZI_AUDIO_RATE> aOscil(SIN8192_DATA);

EventDelay noteDelay;

ADSR<MOZZI_CONTROL_RATE, MOZZI_AUDIO_RATE> envelope;




void setup() {
  Serial.begin(115200);
  aOscil.setFreq(440);
  noteDelay.set(2000);  // 2 second countdown
  envelope.setLevels(255, 0, 0, 0);
  envelope.setTimes(10, 1000, 0, 0);
  startMozzi();
}

void updateControl() {
  envelope.update();
  if (noteDelay.ready()) {
    envelope.noteOn(true);
    noteDelay.start();
  }
}


AudioOutput updateAudio() {
  auto gain = envelope.next();
  return MonoOutput::from16Bit((int)(gain * aOscil.next()));
}



void loop() {
  audioHook();  // required here
}
```

## Result: 
Left is with the PR, Right is before
adsr

### DIFF
--- a/ADSR.h
+++ b/ADSR.h
@@ -72,7 +72,7 @@ private:
 	inline
 	void setPhase(phase * next_phase) {
 		update_step_counter = 0;
-		num_update_steps = next_phase->update_steps;
+		num_update_steps = next_phase->update_steps - LERPS_PER_CONTROL; // so that we stop slightly before, rather than slightly after the end of the phase, avoiding artefacts
 		transition.set(Q8n0_to_Q15n16(next_phase->level),next_phase->lerp_steps);
 		current_phase = next_phase;
 	}
--- a/ADSR.h
+++ b/ADSR.h
@@ -72,7 +72,7 @@ private:
 	inline
 	void setPhase(phase * next_phase) {
 		update_step_counter = 0;
-		num_update_steps = next_phase->update_steps - LERPS_PER_CONTROL; // so that we stop slightly before, rather than slightly after the end of the phase, avoiding artefacts
+		num_update_steps = next_phase->update_steps;
 		transition.set(Q8n0_to_Q15n16(next_phase->level),next_phase->lerp_steps);
 		current_phase = next_phase;
 	}
@@ -91,7 +91,7 @@ private:
 	void setTime(phase * p, unsigned int msec)
 	{
 		p->update_steps = convertMsecToControlUpdateSteps(msec);
-		p->lerp_steps = (long) p->update_steps * LERPS_PER_CONTROL;
+		p->lerp_steps = (long) p->update_steps * (LERPS_PER_CONTROL + 1); // We increase slightly the number of interpolated steps at audio rate to be sure that the transition does not over/under flow because of the approximation made by Line.
 	}
 
 
@@ -99,7 +99,7 @@ private:
 	void setUpdateSteps(phase * p, unsigned int steps)
 	{
 		p->update_steps = steps;
-		p->lerp_steps = (long) steps * LERPS_PER_CONTROL;
+		p->lerp_steps = (long) steps * (LERPS_PER_CONTROL + 1) ;
 	}
 
 


tfry-git:
Your recordings speak a clear language, indeed. I don't think I've quite understood just how this happens, though. Might it need fixing in Line?

tomcombriat:
> Might it need fixing in Line?

Well, not really. For the type of `Line` used by `ADSR`, which is `Q15n16` aka `int32_t`, even if the line slightly overshoot the target value, this is not really a problem, because the outputted values will be very close to the target. For the `unsigned` type, `Line` is using `float` for the calculation of the step size which I bet was done by @sensorium to avoid a more dramatic overflow.

Here, the phenomenon comes from the conversion from the result of `Line` to `unsigned char` [here](https://github.com/sensorium/Mozzi/blob/3500835948cb368a19bac1d42f9a4a240b76d067/ADSR.h#L164). As we are moving to a unsigned type, any substantial overshoot (or undershoot under 0) will lead to an overflow, because of the type rolling back. This small patch just tweaks the ADSR's line to ensure we do not reach that point. Might be worth considering adding another template to ADSR for the return type somewhere in the future though?

I am not sure I am more clear…

tfry-git:
I still don't get, how this is happening. Here's my reasoning (which obviously fails at some point, but where?):

1. ```updateControl()``` (and thus ```envelope.update()```) is called _exactly_ once every ```MOZZI_AUDIO_RATE / MOZZI_CONTROL_RATE``` frames (and thus calls of ```envelope.next()```). This is accurately represented by ```LERPS_PER_CONTROL```.
2. ```envelope.update()``` calls/may call ```Line::set()```. ```envelope.next()``` calls ```Line::next()```. The latter happens exactly the number of times calculated in ```phase->lerp_steps```.
3. ```Line::set()``` essentially just sets the step size (per lerp) to delta/steps (delta is target_level - current_level). This step may be affected by rounding, but rounding always happens towards the smaller (absolute) step_size. Thus ```abs(lerp_steps*Line::step_size)``` is always smaller or equal to ```abs(delta)```.
4. QED, the transition cannot possibly overflow...

Personally, I'm most suspicious about my reasoning in step 3, and it that context about those casts to float in ```Line::set()```. Conceptually, how should that affect the outcome of the division at all, seeing that the result is cast back an integer type, anyway? I do observe in testing that ```static_assert((long) -1.99999995 == -1l);``` fails (it's rounded to -2, instead), but are we actually reaching such a small delta? Anyway, what does happen in the testcase, if you just remove that cast to float?

tomcombriat:
I have to check but I think that
> but rounding always happens towards the smaller (absolute) step_size.

is where the reasoning fails. If I remember, rounding always happens toward the smaller step_size (not in absolute), so negative steps will be too big. But that brings an interesting point I did not realize: this should only for lines with a negative slopes (which explains why I never experienced it the other way around).

tomcombriat:
Update: my bad:
>  If I remember, rounding always happens toward the smaller step_size

Is wrong, you were correct that rounding is done towards 0. I'll investigate more!

tomcombriat:
Okey, I think the problem is more rooted than that… All the way to MozziGuts.hpp…

I think [this line](https://github.com/sensorium/Mozzi/blob/3500835948cb368a19bac1d42f9a4a240b76d067/internal/MozziGuts.hpp#L277) of internal/MozziGuts.hpp which is:

```
update_control_timeout = MOZZI_AUDIO_RATE / control_rate_hz;
```
should be
```
update_control_timeout = MOZZI_AUDIO_RATE / control_rate_hz - 1;
```

## Explanation:
In the current state, there is *one extra* `updateAudio` per `updateControl` than intended: if `MOZZI_AUDIO_RATE  = MOZZI_CONTROL_RATE` this block:
```
inline void advanceControlLoop() {
  if (!update_control_counter) {
    update_control_counter = update_control_timeout;
    updateControl();
#if MOZZI_IS(MOZZI_ANALOG_READ, MOZZI_ANALOG_READ_STANDARD)
    adcStartReadCycle();
#endif
  } else {
    --update_control_counter;
  }
}
```
makes one `updateControl` every **two** times and not every time. (Note that the fix could also be `  if (!--update_control_counter) ` and removing the `else` case.

For ADSR, the discrepancy is small but can manifest like in the example sketch at the top of this PR: `Line` gets exceeded because the number of `next()` is greater than calculated. This error accumulates for long phases, leading to the overflow.

Does this make sense?

tfry-git:
> Does this make sense?

Great find, that makes a whole lot of sense! I like the ```if (!--update_control_counter)```-solution.

(I still wonder, whether the cast to float in ```Line::set()``` is actually called for, but that's not really important)

tomcombriat:
Thanks! Fun that it manifested so far!

> Great find, that makes a whole lot of sense! I like the if (!--update_control_counter)-solution.

And after thinking about it, I think the other `update_control_timeout = MOZZI_AUDIO_RATE / control_rate_hz - 1;` is slightly more efficient as it removes one decrement at `CONTROL_RATE`. Closing this PR, will make a patch soon!

PR #254, opened 2024-05-07
RP2040 on tune

### LINKED ISSUE #251: RP2040 is out of tune
Making a small drone synth to play along, I noticed that a simple setup, based on a RP2040 with PWM output, seems to be out of tune (looks like something similar than #67 ).

Here are some measurements ("Sinewave" sketch, CPU speed: 133MHz (this does not seem to affect):

| Asked frequency (Hz) | Measured frequency (Hz) | Deviation (%) |
|---|---|---|
| 440 | 448 | 1.82 |
| 220 | 224 | 1.82 |
| 1000 | 1017 | 1.7 |
| 2000 | 2035 | 1.75 |
| 10000 | 10173 | 1.73 |

Looks fairly stable, hence the timer is probably off by that amount.

tomcombriat:
Fix #251 

Draft for now, until DMA driven PWM is implemented.


tomcombriat:
Fairly straightforward!

EXTERNAL_TIMED/PWM was initially tested to see that the small patch corrected the detuning.

PWM output has been switched to DMA buffered transfer using PWMAudio. Both mono and stereo have been tested. Crude EXTERNAL_TIMED (with pwm in user sketch) has been tested (not clean at all, the cycle is probably too low, but seems at the correct frequency).

internal/MozziGuts_impl_RP2040.hpp
```diff
@@ -142,30 +143,60 @@ inline void audioOutput(const AudioOutput f) {
...
 uint64_t micros_per_update;
 uint audio_update_alarm_num;
+bool flip_flop=0;
+// NOTE: unfortunately, on the RP2040, alarms can only be set with us resolution. At 32768Hz, we ideally would like to output every 30.51us, so we alternate between outputting after 30us and 31us, using the flip_flop to alternate between these two values,
 
 void audioOutputCallback(uint) {
   do {
     defaultAudioOutput();
-    next_audio_update = delayed_by_us(next_audio_update, micros_per_update);
+    flip_flop = !flip_flop;;
```
tfry-git, line 164: I was rather thinking along the following lines (pseudocode):

```
uint64_t next_audio_update_shifted = 0;
const uint64_t us_persample_shifted = (1000000l << 8) / MOZZI_AUDIO_RATE;

[...]

next_audio_update_shifted += us_per_sample_shifted;
while(hardware_alarm_set_target(alarm_num, delayed_by_us(0, next_audio_update_shifted >> 8));
```

internal/MozziGuts_impl_RP2040.hpp
```diff
@@ -217,9 +248,10 @@ static void startAudio() {
 #if MOZZI_IS(MOZZI_AUDIO_MODE, MOZZI_OUTPUT_PWM)
   // calling analogWrite for the first time will try to init the pwm frequency and range on all pins. We don't want that happening after we've set up our own,
```
tfry-git, line 249: I suppose this comment block is essentially obsolete, then?

internal/MozziGuts_impl_RP2040.hpp
```diff
@@ -231,11 +263,21 @@ static void startAudio() {
 #      error Audio channel pins for stereo or HIFI must be on the same PWM slice (which is the case for the pairs (0,1), (2,3), (4,5), etc. Adjust MOZZI_AUDIO_PIN_1/2 .
 #    endif
   gpio_set_function(MOZZI_AUDIO_PIN_2, GPIO_FUNC_PWM);
+  gpio_set_drive_strength(MOZZI_AUDIO_PIN_2, GPIO_DRIVE_STRENGTH_12MA); // highest we can get*/
+
+  gpio_set_drive_strength(MOZZI_AUDIO_PIN_1, GPIO_DRIVE_STRENGTH_12MA); // highest we can get
+  #  if (MOZZI_AUDIO_CHANNELS > 1)
+#    if ((MOZZI_AUDIO_PIN_1 / 2) != (MOZZI_AUDIO_PIN_1 / 2))
+#      error Audio channel pins for stereo or HIFI must be on the same PWM slice (which is the case for the pairs (0,1), (2,3), (4,5), etc. Adjust MOZZI_AUDIO_PIN_1/2 .
```
tfry-git, line 271: Is this a requirement with PWMAudio, too?

internal/config_checks_rp2040.h
```diff
@@ -85,14 +85,17 @@ MOZZI_CHECK_SUPPORTED(MOZZI_AUDIO_MODE, MOZZI_OUTPUT_EXTERNAL_TIMED, MOZZI_OUTPU
...
+#    define MOZZI_AUDIO_BITS 16 // PWMAudio expects 16bits
 #  endif
 #  if !defined(MOZZI_AUDIO_PIN_1)
 #    define MOZZI_AUDIO_PIN_1 0
 #  endif
 #  if !defined(MOZZI_AUDIO_PIN_2)
 #    define MOZZI_AUDIO_PIN_2 1
 #  endif
+#  define BYPASS_MOZZI_OUTPUT_BUFFER true
+#  define MOZZI_RP2040_BUFFERS 8       // number of DMA buffers used
```
tfry-git, line 97: It may be worth documenting: Is that 8 the result to experimentation, or would other values be expected to work, too?

tfry-git (approved):
Looks good to me (without any testing done). Some comments / questions added.

internal/MozziGuts_impl_RP2040.hpp
tomcombriat: Ah, I understand what you meant now! 

I guess it would even more simpler be: `while (hardware_alarm_set_target(audio_update_alarm_num, next_audio_update_shifted>>8)); // absolute time, so no need for delayed_by_us`

Indeed, that would be more general, I can try. My version works only for this case where we are basically nearly exactly between two values. I have a hard time trying to intuitively figure out which one is faster but yours might be (you: 1 add, 1 shift, me: 2 add, 1 shift). I'll try!

internal/MozziGuts_impl_RP2040.hpp
tomcombriat: Indeeed

internal/MozziGuts_impl_RP2040.hpp
tomcombriat: Yup! (Can't find the reference ATM, but I read that somewhere in PWMAudio)

internal/config_checks_rp2040.h
tomcombriat: I used the same values that the library example (which are the same than for I2S). Other values might be expected to work, but I did not test thoughtfully.

tomcombriat:
PS: this incurs a quite penalty on flash, but is more beneficial for RAM in terms of percentage so I think this is worth it. There is no lack of flash on the RP…

internal/MozziGuts_impl_RP2040.hpp
tomcombriat: @tfry-git What do you think of the new version? Note the needed addition to `startMozzi()`

internal/MozziGuts_impl_RP2040.hpp
```diff
@@ -288,6 +260,7 @@ static void startAudio() {
   micros_per_update = 1000000l / MOZZI_AUDIO_RATE;
   do {
     next_audio_update = make_timeout_time_us(micros_per_update);
+    next_audio_update_shifted = to_us_since_boot(next_audio_update);
```
tfry-git, line 263: I think this is not correct (but wasn't before, either): ```startMozzi()``` may not necessarily be called a time 0. Rather (I think):

```
next_audio_update_shifted = make_timeout_time_us(delayed_by_us(time_us_64(), micros_per_update) << 8;
```

tomcombriat: Are you sure? I am okay to admit that I forgot a shift there, so more like:

```
next_audio_update_shifted = to_us_since_boot(next_audio_update) <<8;
```

But, the `next_audio_update = make_timeout_time_us(micros_per_update);` increment from the time it is called on (does not need to be zeros), but returns an *absolute time*. `delayed_by_us` will return an absolute time, in the future already, and the `make_timeout_time_us` will give a "timestamp a number of microseconds from the current time.", hence a absolute time, starting at the current time and delayed by (current_time + micros_per_update), returned by the `delayed_by_us`. I might be wrong though because I am not sure I understand all the `absolute_time_t` correctly. The shift is missing though.

tfry-git: You are right, I got confused with relative and absolute timestamps!

PR #260, opened 2024-06-18
Fix/audio input

### LINKED ISSUE #259: AudioInput does not compile with Mozzi 2.0
(Initially reported on the forum)

Example `Audio_Input` does not compile with Mozzi 2.0. Looks like a simple typo, will provide a PR in a minute.

tomcombriat:
Fix #259


internal/config_checks_generic.h
```diff
@@ -126,7 +126,7 @@ MOZZI_CHECK_POW2(MOZZI_CONTROL_RATE)
 #error "MOZZI_AUDIO_INPUT depends on MOZZI_ANALOG_READ option"
 #endif
 
-#if !MOZZI_IS(MOZZI_AUDIO_INPUT, MOZZI_AUDIO_INPUT_NONE) && defined(MOZZI_AUDIO_INPUT_PIN)
+#if MOZZI_IS(MOZZI_AUDIO_INPUT, MOZZI_AUDIO_INPUT_NONE) && defined(MOZZI_AUDIO_INPUT_PIN)
```
tfry-git, line 129: This one was correct before: Shows a warning, if MOZZI_AUDIO_INPUT_PIN is defined without MOZZI_AUDIO_INPUT

tfry-git (approved):
I think the change in config_checks_generic.h is wrong, but otherwise: go ahead!

internal/config_checks_generic.h
tomcombriat: Are you sure?
 
The old one is: If Mozzi_input is Not None (hence, it is defined) and a pin is defined, shows an error. This gives a warning when compiling the Audio_Input example.

The new one is: If Mozzi_input is None and a pin is defined, shows an error.

tfry-git: You are right!

PR #277, opened 2024-09-22
Fixed compilation for Teensy2

tomcombriat:
Looks like a typo, but do not have the hardware to try…

### DIFF
--- a/internal/MozziGuts_impl_AVR.hpp
+++ b/internal/MozziGuts_impl_AVR.hpp
@@ -39,7 +39,7 @@ uint8_t adcPinToChannelNum(uint8_t pin) {
 	// 0, 1, 4, 5, 6, 7, 13, 12, 11, 10, 9, 8
 	0, 1, 4, 5, 6, 7, 13, 12, 11, 10, 9, 8, 10, 11, 12, 13, 7, 6, 5, 4, 1, 0, 8
 	};
-	pin = pgm_read_byte(adc_mapping + (P));
+	pin = pgm_read_byte(adc_mapping + (pin));
 #  else
 	pin = analogPinToChannel(pin);
 #  endif


tfry-git:
Yes, clearly a typo. But no, I don't have a Teensy board, either.

@sensorium Could you enable the permission  "Send write tokens to workflows from fork pull requests" as per https://github.com/arduino/report-size-deltas?tab=readme-ov-file#run-from-a-scheduled-workflow ?

tfry-git:
And here's the offending commit introducing the error: https://github.com/sensorium/Mozzi/commit/51ae09a918a81009a38d47996a79cc179dc77bd3

tomcombriat:
Alright. If you agree I'm in favour of merging, and worse case changing it again if it does not work as intended: that only affects Teensy2 and at least it compiles now…

sensorium:
> permission "Send write tokens to workflows from fork pull requests" as per https://github.com/arduino/report-size-deltas?tab=readme-ov-file#run-from-a-scheduled-workflow ?

Hi @tfry-git , I followed the link from that page to "Send write tokens to workflows from fork pull requests", and that option seems to be only available for private repos?  I couldn't find that permission in my settings under actions/general... should I look somewhere else?

tfry-git:
> > permission "Send write tokens to workflows from fork pull requests" as per https://github.com/arduino/report-size-deltas?tab=readme-ov-file#run-from-a-scheduled-workflow ?
> 
> Hi @tfry-git , I followed the link from that page to "Send write tokens to workflows from fork pull requests", and that option seems to be only available for private repos? I couldn't find that permission in my settings under actions/general... should I look somewhere else?

Hm, I can't seem to find the option, either. Sorry for sending you to a dead end.

Issue #280, opened 2024-10-11
Should all tables be made "symetrical"?


tomcombriat:
I have be wondering about that for some time and #279 re-triggered this question in my mind, and this issue is to have others' opinions

Most of legacy tables of Mozzi, which are of `int8_t` have a maximum value of 127 and a minimal value of -128, taking the full range of the type. 

Wouldn't it preferable to have them ranging from -127 to 127?

### Pros:

 - they would be naturally centered around 0, and not around -0.5
 - negating them (ie, inverting the phase) would be easier (as of now, negating of these poses problems as -(-128) cannot be represented in `int8_t`, leading to an overflow.

### Cons:

- slightly less range
- might have a very slight effect on existing sketch (this is probably very mild though).

Note that all the band_limited tables are actually already symmetric. 

Any thoughts?

tomcombriat:
Hi again,
I am just starting to dive into this and I am actually interested in an external opinion.
Looking more closely at the tables, a fairly good chunck of them are marked as `int8_t` type, even though the internal data are `uint8_t`. Note that `Oscil` only supports `uint8_t` as of now… Note also that even the band limited tables I generated for non-aliased sound are also wrongly marked… 
I am a bit tempted to rename all of that to make it coherent, but that might have big consequences for users… So what is the best strategy in your opinion?

 - let it be like it is as of now, knowing that `uint8_t` and `int8_t` refer to the same thing. They can still be rescaled. 
 - changing them all, deprecating the old one with a warning that they might be deleted at some point and pointing to the correct one.

I am probably for option 1, but would like to have other opinions.

---

Also, I realized that `cos` tables are mostly describing `-cos`. Not a very big deal IMO but while at it I could put them back at the correct phase. What do you think?

eclab:
I am confused by this comment:  as far as I know, oscil accepts int8_t, not uint8_t in its constructor.  See https://sensorium.github.io/Mozzi/doc/html/class_oscil.html

tomcombriat:
That's a typo, this should read

>  only supports `uint8_t` as of now…

Edited in the OP

tomcombriat:
@eclab Indeed, I do not know my mind shifted to: "These tables should be `uint8_t`. "

tomcombriat:
@eclab Any opinion on #284 ?

eclab:
[...]

PR #284, opened 2024-10-21
Fix/tables

### LINKED ISSUE #279: 8-bit Sine tables to not describe sines, but rather negative cosines
Most of the sine tables do not start at the origin, but rather at -128.  This is bad because it can cause a click if the wave is played raw.  They should start at 0, then increase from there, then back down to -128.

tomcombriat:
Fix #279


tomcombriat:
Alright, this also fixes #280. Here was my method:

 - for "mathematical" tables, like saw, tri, cos and sin, I recomputed them from scratch, for instance:

``` python
length = 2048

dat = []
out = open("out.txt","w")
t = np.linspace(0,2*np.pi,num=length,endpoint=False)
out.write("\t")
for i in range(len(t)):
    #out.write(str(round(127*np.sin(t[i])))+", ") # for the sin
    #out.write(str(round(-127+127*2.*i/(length-1)))+", ")  # for the saw
    # for the tri
    if (i<=length/4):
        dat.append(i*127./length*4)
    elif (i<=length*3/4):
        dat.append(127-(i-length/4)*(127*2)/(length/2))
    else:
        dat.append(-127.+127.*(i-length*3/4)/length*4)
```
for the `sin`.

 - tables that were not touching `-128` were left as they were.
 - other special tables (noises, scaler, waveshaper and actually square), I tried to touch them as little as possible, just rescaling slightly to get rid of the -128:

```python
if (np.amin(data)==-128):
    data = (data-np.amin(data)) * (np.amax(data)-(np.amin(data)+1)) / (np.amax(data) - np.amin(data)) + np.amin(data)+1
```
where `data` is the original content of the table. This rescale between the original maximum and -127 if the original table was going to -128.
 Note that this does not ensure that tables have a 0-mean, but that was not necessary the case to begin with. I think it is better to keep the tables as close as possible for their original which this is achieving.

What do you think?

tomcombriat:
For reproducibility purposes, I added the script I used to regenerate the "standard" tables. Plus the tables automatically generated.

tomcombriat:
Any opinion? Can I merge this?

sensorium:
Great work Tom.  OK by me to merge.

PR #286, opened 2024-11-03
Add a type template parameter to `AudioDelayFeedback`

tomcombriat:
Work in progress! 

The idea would be to be able to have types accepted by `AudioDelayFeedback`, exactly the way it is done for `AudioDelay`.

### DIFF
--- a/AudioDelayFeedback.h
+++ b/AudioDelayFeedback.h
@@ -16,6 +16,7 @@
 
 #include "mozzi_utils.h"
 #include "meta.h"
+#include "IntegerType.h"
 
 enum interpolation_types {LINEAR,ALLPASS};
 
@@ -32,7 +33,7 @@ than a plain AudioDelay, but allows for more dramatic effects with feedback.
 @tparam INTERP_TYPE a choice of LINEAR (default) or ALLPASS interpolation.  LINEAR is better
 for sweeping delay times, ALLPASS may be better for reverb-like effects.
 */
-template <uint16_t NUM_BUFFER_SAMPLES, int8_t INTERP_TYPE = LINEAR>
+template <uint16_t NUM_BUFFER_SAMPLES, int8_t INTERP_TYPE = LINEAR, typename su=int8_t>
 class AudioDelayFeedback
 {
 
@@ -68,7 +69,7 @@ public:
 	@note slower than next(int8_t input, uint16_t delaytime_cells)
 	*/
 	inline
-	int16_t next(int8_t input)
+	typename IntegerType<sizeof(su)+sizeof(su)>::signed_type next(su input)
 	{
 		// chooses a different next() function depending on whether the
 		// the template parameter is LINEAR(default if none provided) or ALLPASS.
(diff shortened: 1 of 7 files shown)


tomcombriat:
Untested but I think this should be it. I am leaving some comments on things I am not sure about. Will do a more consistent message once testing is done.

Cheers,

AudioDelayFeedback.h
```diff
@@ -290,19 +299,24 @@ class AudioDelayFeedback
 		= coeff * (in-last_out) + last_in
 		*/
 		//setPin13High();
+
+	  /* I think these should **not** be static 
 		static int8_t last_in;
```
tomcombriat, line 304: These two were static originally. I think that prevents this class from being used as several instances, as they would share `last_in` and `last_out`, but maybe I am being mistaken?

sensorium: I'm not sure, Tom.  Definitely worth checking.
They're inside a function definition and in that context they persist between function calls, but are inaccessible from outside.  They can also have a speed benefit if cached locally.  
If they were static variables in the class outside the function, you're right, they would be shared between all objects of that class and also accessible from outside the class with ::, without even needing an object to be instantiated.  (I just googled this yesterday for something else - it's probably worth checking).  
Also, the example AudioDelayFeedbackX2 seems to work...?

sensorium: Aha, you're right (as usual).
It will be interesting to hear what the Mozzi example sounds like when it works properly!

tomcombriat: Will test nevertheless. I was initially confused not to see any impact on the reports by actually these examples are not compiled in the autotests.

tomcombriat:
Ok, the `static` does not seem to make much difference in the example, but I did not try a mock up example yet to see how the Arduino compiler does understand this case.

I corrected a gross mistake in my use of `constrain` and starts to sound good. I am still investigating an overflow in the example `AudioDelayFeedbackX2` but which also seems to appear in the master branch (?).

tomcombriat:
(Small update on this saga, actually, the example itself is having an overflow. This [line](https://github.com/sensorium/Mozzi/blob/93a51c0c04d4f8094131dab85cfd568875864f1b/examples/09.Delays/AudioDelayFeedbackX2/AudioDelayFeedbackX2.ino#L59)
```
del_samps1 = 64+kDelSamps1.next();
```
have the right term going to the negative. Somehow, I guess this would be reinterpreted right on 8bitters (haven't tested) but this does not sound like the example on 32-bitters. Next time I put myself to it (a bit late now), will try to change that to `del_samps1 = 128+kDelSamps1.next();` to see if that is any more closer to the initial closing.

Sorry for the spam, that's also for me remembering…

tomcombriat:
Ready for me!

I corrected a few delay examples using `char` as per #287 , and one example which had an overflow. Docs will need to be regenerated before the next release.

tomcombriat:
I will merge this, from testing it seems that the default version is identical to the previous one (except for the two `static` variables) and the version with `int16_t` also works well (it eats a lot of RAM though). 

I am using this class quite a bit these days so if anything wrong went through that will be spotted quite fast. Planning to release in a couple of weeks!

PR #295, opened 2024-12-05
Fixes for flot2mozzi.py script

leobel96:


leobel96:
As mentioned in the commit message:
The script allows to create mozzi tables from raw files and it's supposed to replace all the "*2mozzy.py" scripts thanks to its high configurability (it also supports int16/int32 tables in case they'll be supported in the future)

tomcombriat:
Haven't try but looks good to me at first glance!

One thing that might be worth adding as an option would be to allow to make the table "symmetrical" (see #280 ), which, I think Audacity is not doing per default but which can be made easily with your script?

leobel96:
> One thing that might be worth adding as an option would be to allow to make the table "symmetrical"

Uhm, I guess that you are talking about Oscil wavetables here, not Sample ones, right?

tomcombriat:
Principally yes, but can also be relevant for samples? (This allows to negate them).

leobel96:
And with "symmetrical" you don't actually mean that the first half of the table has to be identical to the specular of the second part, right? You only mean that the last value has to be identical to the first one to not have a gap and avoid weird artifacts. So you just want me to append a tail of cells to the table that gradually connects the last cell to the first one. Did I understand everything correctly?

tomcombriat:
Ah, sorry for the confusion, I think symmetrical I maybe not the right word. By that I mean more a symmetrical range, so that the samples go from -127 to 127 and not to 128. That does not the sound but that kind simplify things sometimes... That's maybe nitpicking...

For the start and end point, I think it makes sense that they start at 0 to avoid a click, but I guess that's more the job of the one cutting the samples than your script....

leobel96:
> so that the samples go from -127 to 127 and not to 128

Uhm, how can they go to 128 if they are `int8_t` 🤔? Did you mean -128? So you mean symmetrical respect to the X axis in a way that the minimum negative value is equal to the maximum positive one in absolute value?

tomcombriat:
Yes, sorry for the typo and the confusion.
I guess Samples do not have to span the full range actually, but if it is easy to make the script so that -128 is not present that might be useful.

leobel96:
So is the whole thing just a matter of replacing all the -128 with -127? That's it? 😂

tomcombriat:
Well, that's one way to do it ;). But a cleaner way would be to renormalize the whole signal in that range (to avoid saturation/hard clip) around -127. But I actually think that it is probably nearly equivalent when starting with 8bits…

(For instance, you can check how I did it for some of the existing tables [here](https://github.com/sensorium/Mozzi/pull/284#issuecomment-2427712078) (second code snip).

leobel96:
Ah! This makes sense, your snippet makes everything more clear. I can add this bit in the next few days and amend the commit above

tomcombriat:
Sometimes a bit of code is better than a full speech ;). Sorry for the confusion

leobel96:
Hey! I finally found some time to add that symmetrical part. Do you mind testing the script and, if it works as expected, merge it? @tomcombriat

tomcombriat:
Nice! I'll have a look at it very soon!

Thanks for this

extras/python/audio2mozzi.py
```diff
@@ -0,0 +1,167 @@
...
+            )
+            # Mega2560 boards won't upload if there is 33, 33, 33 in the array, so dither the 3rd 33 if there is one
+            if value == 33:
+                cnt_33 += 1
+                if cnt_33 == 3:
+                    out_values.append(random.choice((32, 34)))
+                    cnt_33 = 0
+            else:
+                cnt_33 = 0
+        if args.make_symmetrical:
```
tomcombriat, line 95: I think this block should read:
``` python
        if args.make_symmetrical:
            value_to_remove = -max_output_value 
            if value_to_remove in out_values:
                in_min = value_to_remove
                in_max = max_output_value
                out_max = max_output_value
                out_min = -max_output_value+1
                out_values = (int((value - in_min) * (out_max - out_min) / (in_max - in_min) + out_min) for value in out_values)
```

in the current version: `value_to_remove = -max_output_value - 1` is -128-1=-129 for 8bits, hence the test is never passed. `out_min` should be set to -127 in that case to return the correct rounded value. Also note the conversion back to `int` :)

extras/python/audio2mozzi.py
```diff
@@ -0,0 +1,167 @@
...
+
+        table = f"CONSTTABLE_STORAGE(int{args.output_bits}_t) {tablename}_DATA [] = {{"
+        max_output_value = 2 << (args.output_bits - 2)  # Halved because signed
+        out_values = []
+        for value in in_values:
+            cnt_33 = 0
+            out_values.append(
+                math.trunc((value * max_output_value) + 0.5)
+                if args.input_encoding == "float"
+                else value
```
tomcombriat, line 85: Shouldn't some kind of normalization happen here in case of 16bits as input? From my tests, inputting a 16bits and expecting an 8bits output produces some way out of range values.

extras/python/audio2mozzi.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser(
+        description="Script for converting a raw audio file to a Mozzi table",
+    )
+    parser.add_argument(
+        "-e",
+        "--input-encoding",
+        choices=("float", "int"),
```
tomcombriat, line 122: In float mode, with default number of output bits, some values reach 128 (which do not fit). Maybe you should discard my comment on the make symmetric but subtract -1 to the max output value?

extras/python/audio2mozzi.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python3
+"""
+For converting raw audio files to Mozzi table
```
tomcombriat, line 3: Maybe add some copyright info (you or Mozzi team and you depending on how much you drew from the original script) and a license (LGPL and GPL are already in Mozzi)?

extras/python/audio2mozzi.py
```diff
@@ -0,0 +1,167 @@
...
+        fout.write(f"#define {tablename}_NUM_CELLS {len(in_values)}\n")
+        fout.write(f"#define {tablename}_SAMPLERATE {args.sample_rate}\n\n")
+
+        table = f"CONSTTABLE_STORAGE(int{args.output_bits}_t) {tablename}_DATA [] = {{"
+        max_output_value = 2 << (args.output_bits - 2)  # Halved because signed
+        out_values = []
+        for value in in_values:
+            cnt_33 = 0
+            out_values.append(
+                math.trunc((value * max_output_value) + 0.5)
```
tomcombriat, line 83: In conjunction with my other comments on the output range, might be easier to use `round` instead of `trunc`?

tomcombriat (changes requested):

extras/python/audio2mozzi.py
tomcombriat: (Note that I did not look at the original script, where some of the rough edges I found might already be ;).) The script is very easy to use (and made me discovered `ArgumentParser`… I was still doing that manually…)

extras/python/audio2mozzi.py
```diff
@@ -0,0 +1,167 @@
...
+                out_min = -max_output_value
+                out_values = (((value - in_min) * (out_max - out_min) / (in_max - in_min) + out_min) for value in out_values)
+
+        table += ", ".join(map(str, out_values))
+        table += "};"
+        table = textwrap.fill(table, 80)
+        fout.write(table)
+        fout.write("\n\n")
+        fout.write(f"#endif /* {tablename}_H_ */\n")
+        print(f"wrote {table} to {output_path}")
```
tomcombriat, line 110: I would personally remove the `print` statements that, IMO, clutters the terminal.

tomcombriat: (or maybe as a `--verbose` argument?

leobel96: Okay, added a `--verbose` argument for this

extras/python/audio2mozzi.py
leobel96: uhm, I didn't really understand this. Do you mean that the output of the script reaches 128? That's not expected

extras/python/audio2mozzi.py
leobel96: Everything should be nicer now 🙂

extras/python/audio2mozzi.py
leobel96: Yeah, you are right, I didn't test it properly. Using some help from Numpy 😄

extras/python/audio2mozzi.py
leobel96: Yeah, this `trunc` thing was already present in the original script and confused me too. I wanted to keep this script as close as possible to the original. I think that round would work better too. Fixed in the new version.

extras/python/audio2mozzi.py
leobel96: Done!

tomcombriat:
Thanks! Give me a few days to test it :)

extras/python/audio2mozzi.py
```diff
@@ -0,0 +1,211 @@
...
+        if args.output_bits == 8:
+            output_dtype = np.int8
+        elif args.output_bits == 16:
+            output_dtype = np.int16
+        elif args.output_bits == 32:
+            output_dtype = np.int32
+        else:
+            raise Exception("Unrecognised --input-bits value")
+
+        output_array = input_array.astype(output_dtype)
```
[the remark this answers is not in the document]
tomcombriat: Adding on this, I see now that your mapping is not used when make_symmetrical is false. `astype` does not change the values, but just reinterpret them to the new type (100 as int8 will stay 100 as int16).

tomcombriat (changes requested):
Sorry for the delay!

Works like a charm when converting to the same number of bits, but I think still fails a bit (pun unintended) when having different number of bits between input and output (that should work right?):
 - when scaling down (from 16 to 8 say) the output is very weird, looks like an overflow from the l.120.
 - when scaling up (from 8 to 16 say), the values are not changed (they stay in the 8 bits range)
 - when using --make-symmetrical, the values in the output files are floating points (even when saying 8 bits as an output)..

Here are the files I use for testing (simple sine wave, generated in Audacity, 1s at 8192sampling rate):
[raws.zip](https://github.com/user-attachments/files/18270875/raws.zip)

Issue #297, opened 2024-12-11
ESP32 Dev Module stops working after including Mozzi.h


sensorium:
This sketch works:

        //#include <Mozzi.h>
        
        void setup(){
          pinMode(22, OUTPUT);
          digitalWrite(22, HIGH);
        }
        
        void loop(){
        }

But this doesn't:

        #include <Mozzi.h>
        
        void setup(){
          pinMode(22, OUTPUT);
          digitalWrite(22, HIGH);
        }
        
        void loop(){
        }

tomcombriat:
Just received some ESP32! Will try here if I can reproduce.

tomcombriat:
Ok, I can confirm this. I also get a lot of "Deprecated" warnings on compilation. Looks like the ESP32 port needs an update!

tomcombriat:
Ok, this seems to be exactly the same than #261 : a conflict because we are using the legacy i2s implementation. I do not really understand why the conflict as we are only using the legacy of all drivers but…

Probably worth moving everything to the new API, I'll give it a try in the branch fix/esp32, but if @tfry-git or @dietervandoren which I think are more accustomed with the platform want to give it a try, they should feel welcome :)

cammycam:
Ahh thank you, very much appreciate you updating for esp32!

PR #301, opened 2025-01-06
Fix ESP32 port

### LINKED ISSUE #278: The ESP32 S3 does not have a built-in DAC, but it has been identified as an ESP32 device and cannot use PWM analog output. Is there any way to solve the problem of using ESP32S3 normally without external or internal DAC
The ESP32 S3 does not have a built-in DAC, but it has been identified as an ESP32 device and cannot use PWM analog output. Is there any way to solve the problem of using ESP32S3 normally without external or internal DAC

tomcombriat:
## Note:

I have done everything using the ESP-IDF documentation (https://docs.espressif.com/projects/esp-idf/en/v5.4/esp32/index.html) but not the Arduino-esp32 one (https://docs.espressif.com/projects/arduino-esp32/en/latest/api/i2s.html).

## ~Work in progress~ Done (?)

The goal is to change the port for the ESP32 to the new version of the API (v.5). Below the state of the work.

**Note:** only tested on ESP32-WROOM-32D (which have a native DAC) and ESP32-S3 (which does not).

---

### Fixed

Can be tested with, as a config:
```c++
#include<MozziConfigValues.h>
#define MOZZI_AUDIO_MODE MOZZI_OUTPUT_EXTERNAL_TIMED 
#define MOZZI_AUDIO_BITS 8 
```

and as an `audioOutput()`:
```c++
void audioOutput(const AudioOutput f) // f is a structure containing both channels
{
  dacWrite(25,f.l()+MOZZI_AUDIO_BIAS);
}
```

### ~Compiles and run (does not panic) but no sound~ Fixed (see below)

Still resists me… I tried to switch to a DMA written DAC as per the [docs](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/dac.html#_CPPv427dac_continuous_new_channelsPK23dac_continuous_config_tP23dac_continuous_handle_t) but still got some error message from the ESP… If anyone has an idea… Worst case, something like the `dacWrite` used in the example above could do the job, but that's not really satisfying…

Note: https://github.com/espressif/arduino-esp32/issues/10851

Mozzi runs, `updateControl` is called, i2s writing function is called… 

~**Edit**: I think I'll do the workaround described for the timed_output using `dacWrite` in the near future, and will change when espressif gives some feedback (or I find the problem myself…)~
I did the workaround above. Works well but probably not the most efficient. Will probably evolves but that works.

**Default mode for ESP32**

### Fixed

Default config to plain format (**different than before**) to be coherent with the others I2S ports (especially RP2040). Mode can be further changed to LSBJ (PT8211) and lines phases can also be edited.
Works in MONO and STEREO for the PT8211 (do not have the DAC to test plain, or at least not in super close reach).

For some reasons, the signals are way cleaner when outputting in stereo, even in mono… And it was the case before… Digital observation of the signals does not reveal anything strange but, for the sake of clean output, the I2S is internally configured as STEREO.

### Fixed

Took me a bit of time because I never used PDM… But seems to work fine now!

## `PWM`
New addition, mainly to support the ESP32 chips that do not have a DAC on-board. A bit fancier than a `analogWrite` but not much. Default mode automatic selection is not yet implemented (and this is not tested on other boards yet…). This will fix #278 .
Default pins are 15 and 16.

**Default mode for all other boards (ESP32-XX)**


amiel:
Thank you so much for working on this!!!!!! I am using MOZZI_OUTPUT_I2S_DAC with a PT8211 in a side project and will test this out when I get a chance, although it won't be a for a few weeks.

tomcombriat:
Alright, sorry for the delay since the original issues (but I did not used to own an ESP32 ;) but this should update the ESP32 port to the new API. 

Seems to work well here (tested on both an ESP32 and ESP32-S3) but testing is always welcome, so mentioning thereafter people that might be concerned: @sensorium , @kicetwigs , @thomasfredericks , @MatteoHMarangoni  , @1183747933 , @dietervandoren , @amiel .

amiel:
Thank you so much!!!!!

As of yesterday, this was working well for me using an esp32 (just plain original esp32) with:

```c
#define MOZZI_AUDIO_MODE MOZZI_OUTPUT_I2S_DAC
// and one custom pin
#define MOZZI_I2S_PIN_WS 25
```

I'm using headphones with the PT8211, which is not recommended, so it's hard to say if the sound is good.

I will test again tomorrow with the latest changes.

tomcombriat:
@amiel Thanks for this! I think, for a PT8211, you should also have:

``` c++
#define MOZZI_I2S_FORMAT MOZZI_I2S_FORMAT_LSBJ
```

amiel:
Ohhhh.Thank you, I'll try that out!

amiel:
Have you ever used a MAX98357A? That was my initial attempt and I never got it to sound right (but the PT8211 worked out-of-the-box with mozzi 1). I'm wondering if maybe I just needed to do something like `MOZZI_I2S_FORMAT_LSBJ`.

tomcombriat:
@amiel 

>  That was my initial attempt and I never got it to sound right (but the PT8211 worked out-of-the-box with mozzi 1)

Was that with an ESP32?

There are two (at least…) formats to I2S: standard, and LSBJ (and variants on each -_-), PT8211 is LSBJ but most DAC are on the standard. If I got the previous port okay, the I2S was configured for LSBJ and not modifiable to the other format which would explain what you observed. In order to keep coherency with the other I2S ports (especially RP2040) I changed the default format to plain (which should work with the [MAX](https://www.analog.com/media/en/technical-documentation/data-sheets/MAX98357A-MAX98357B.pdf):

> the MAX98357A follows standard I2S timing by allowing a delay of one BCLK cycle after the LRCLK transition before the beginning of a new data word

I have not tested though (I kind of remember now that I might have one around…)

amiel:
Yes, it was with an ESP32. 

>  the I2S was configured for LSBJ and not modifiable to the other format which would explain what you observed.

This is really good to know. I will try the MAX98357A again, as I would prefer to use that for my project anyway. I'll let you know how it goes if that would be helpful.

amiel:
FYI: This is working perfectly with the MAX98357A (no LSBJ). Thank you so much!

tomcombriat:
@amiel Thanks for the feedback!

tomcombriat:
@amiel Did you try in stereo with this DAC? I think it was not completely correct then (fixed now).

sensorium:
Hey Tom, thanks so much for wading through all that!  I'll try to test with the board I have here (SparkFun ESP32 Qwiic Pro Mini) in the next few days.

tomcombriat:
Thanks @sensorium ! I tried to be thoughtful in testing so if that is positive on your side I think we can merge (the ESP32 port is anyway completely broken as of now so can't be worse…). And maybe have a new release?

sensorium:
Hi Tom,
I can't get a sound!  

This board has an internal DAC.  What's the minimal test to check if it's working?I tried the fix/esp branch with the sinewave example, first without adding anything, then:
 
    #include "MozziConfigValues.h" 
    #define MOZZI_AUDIO_MODE MOZZI_OUTPUT_INTERNAL_DAC
and later:
    #define MOZZI_AUDIO_MODE MOZZI_OUTPUT_PWM

I listened on pins 25, 26 for the DAC, and 15, 16 for PWM.  (Actually every pin, each time).  What am I missing?

tomcombriat:
Hi @sensorium 

Arf, this would have been too easy it seems!

I unfortunately don't have this board to test but this board should fall in the "standard ESP32" box. There are nevertheless a few things than can be tried:

 - what core are you using? And which version? (I am on the esp32 by Espressif Systems, v3.1.1). Which specific board do you select in the menu?
 - does the code in #297 with the blinking works?
 - while running Mozzi, and after adding a `Serial.begin(115200);` in your setup, does the board says anything on the Serial monitor?

tomcombriat:
Also have you tried any of the other modes (just to narrow down what is wrong)? PWM, PDM and I2S (but harder to test without a DAC) should work on that board also.

sensorium:
Hi Tom!
Bad news, sorry.  The basic Arduino blink example works, until I add `#include <Mozzi.h>`
Then it uploads, but doesn't blink.  The LED looks very dimly on, but no blink.

I'm using esp32 by Espressif Systems, v3.1.1, and selecting "esp32 Dev Module" as advised on the Sparkfun page for this board (reminder it's the SparkFun ESP32 Qwiic Pro Mini).
Or when you ask what core I'm using, do you mean in the Arduino menu?  It's set to core 1 both for "Events" and "Arduino".  I don't think I've changed it, that's just how it was set.  Should that be different?
Adding `Serial.begin(115200);` doesn't print out anything.
I tried all the output modes.  I think some had different effects like total silence or glitchy noise, but nothing like a sine wave.

Which board are you using?  I'd be happy to try that, stop bugging you, and move on.

tomcombriat:
Hi @sensorium , sorry for the delay…

Hum, hard to know what is wrong :/… Re-tested here after an update (espressif 3.1.3) and with both the "esp32 Dev Module" or "ESP32-WROOM-DA module" selected and it works with both with an unmodified sinewave example, sound on pin 25.

 For the record, I am using a development from a big Chinese website, exact chip is an ESP32-WROOM-32D which should be compatible with yours. I have also tested on a "ESP32-S3-WROOM-1" (the I2S port for the latter appears to be broken with 3.1.3, but not on 3.1.1… Never ending story…).


On which pin did you try the PWM mode? Can't find the GPIO 15/16, or I got the board wrong: https://docs.sparkfun.com/SparkFun_ESP32_Qwiic_Pro_Mini/hardware_overview/#gpio

sensorium:
It works!  Seems weird, but after the update to espressif 3.1.3, I now have a sine wave on pin 25, with both modules as you mentioned.  Thanks Tom!

Also, yes you're right, no gpio 15 or 16 on this board - I think I was confused partly because the datasheet for the ESP32­ PICO ­MINI­ 02 shows io 25 and 26 on pins 15 and 16 of the chip.  Takes me a while to get familiar with stuff.

Well, apologies for the hold-ups, and thanks for your patience.  I'm happy for the merge when you and @tfry-git are ready.

tomcombriat:
> It works! 

Hooray! Maybe a small bug fix from Espressif's side that got solved very recently.

> Well, apologies for the hold-ups, and thanks for your patience. 

Thanks to you for testing! If you do not object I will merge later today and make a new release!

amiel:
Another test case here. I haven't actually wired it up yet but my collaborator says this is working with a [CJMCU-1334 DAC Module](https://www.amazon.com/dp/B0CPSP65VY?ref=ppx_yo2ov_dt_b_fed_asin_title).

tomcombriat:
Merging. Thanks for @amiel and @sensorium for testing!

PR #320, opened 2026-01-02
Remove old include<i2s.h> on MozziGuts_impl_ESP8266.hpp to avoid build failure

ncravino:
The include of "i2s.h" (lower case) will make compilation fail on esp8266. 

The necessary include "I2S.h" (upper case) is already on line 43.

Removing the i2s.h (lower case) will build and work with i2s for both PDM and DAC cases. 

For MOZZI_AUDIO_MODE = MOZZI_OUTPUT_PDM_VIA_I2S:
- Tested with WeMos D1 using RX (I2S output) 

For MOZZI_AUDIO_MODE = MOZZI_OUTPUT_I2S_DAC:
- Tested with Wemos D1 + PCM5102 I2S board, using IO15->BC(L)K, RX->DIN, IO02->L(R)CK.


tomcombriat:
Hello,
Thanks for this! I am happy to merge shortly if @sensorium and @tfry-git do not have anything against that.

PR #328, opened 2026-04-29
Add wav2mozzi.py to convert .WAV files to Mozzi headers

positron96:
Hi. I've found using char2mozzi with raw files a bit cumbersome, so I made a similar script that takes WAV files.

It reads bitness, sample format, and sample rate from the WAV header automatically.
Supports 8-bit unsigned, 16-bit signed, 24-bit signed, and 32-bit signed PCM WAV files, as well as 32-bit IEEE float WAV files (samples in -1.0..1.0 range).
Everything is converted to int8 C array, format taken from char2mozzi.
If file is stereo, only first channel is used.

I've tested it on some of my files, but will test it more extensively if there is interest.


tomcombriat:
Hi,
Thanks for this! There was some previous attempts to change this old script (see #295 for instance), if you are curious.
I will try to have a look at it soon, do you still expect to make some changes?

positron96:
I am gradually testing it with some of the files I have, but so far it works, so no changes in this regard. 

Regarding the PR you referenced, I can add similar symmetrical int8_t conversion here too. 
Also, it mentions output format different from int8_t, how important is it? I had that initially (with int16 output), but then saw that only int8 is currently supported by the library, so removed it to simplify code.

tomcombriat:
Sorry for the late reply.

> Regarding the PR you referenced, I can add similar symmetrical int8_t conversion here too.

I think this would be cool. After some debates, there was some consensus that symmetrical waveforms are usually preferable as negating them can be done without risking an overflow if the type is not changed before.

Regarding other types, indeed only `int8_t` is supported by Mozzi as of now. However, there have been some work to add 16 bits oscillator support (see for instance #133 ). This has not been pursued further as of now, but this might get some light in the future especially as Mozzi is supporting more powerful, 32 bits platforms. So I would say that, if you already have the code, and is not too much work, this is worth keeping, maybe with a warning.

positron96:
I've added 2 options: output bitwidth selector (8/16 bits, 8 by default), and symmetric mode (on by default)

tomcombriat:
Alright, looks very good to me! Tried it with samples of different bitness and works like a charm. 
A few comments that you may address or not before merging this:

 - I had to change the sheband to `python3` instead of `python` for it to work here as a program. That might be my python installation though. (I am amazed that this script is without dependencies, not even numpy…).
 - You might want to add some copyright info?
 - I see that you are dithering around 33 value, just like char2mozzi.py was doing. I am not sure if this is still valid but do not have the board at hand to test right now. This can always be changed later.
 - As this is supposed to replace char2mozzi (I think), I wonder if the instructions at the top of this file should not be added to this new script, as the workflow is basically the same, except way more powerful. That would allow us to deprecate char2mozzi and ultimately remove it. What do you think?
Thanks for that!

positron96:
Yeah, good points! I used windows for development, so haven't tested the shebang, now this should be fixed (and the file is also made executable on linux). 
Documentation is rearranged so that it's printed by `wav2mozzi.py --help`, with Audacity bits copied from char2mozzi.

tomcombriat:
Looks very good, thanks. Merging, this will be part of the next release.

PR #330, opened 2026-06-22
Add pause functionality to Sample

### LINKED ISSUE #329: FR: stop sample playback
Doesn't seem that people have asked this before, but is there a way to _not_ play a sample on boot? 

I'm trying to use Mozzi to play sound effects on a diorama, for this I need to start samples based on events/timers, etc.
`Sample.start` works well for this as it resets sample position to 0 (and it will effectively start playing from the beginning), but also all the samples start playing right at the start of the sketch!

If there isn't a way to do this, this is a feature request to add a "Sample.stop" method. After calling it, `Sample.next()` should return 0. Quick look at the source shows that it should be possible to implement it by moving sample position to the end of the sample, this will also support proper functioning of isPlaying function, etc. If the implementation is that simple, I can provide a PR.

positron96:
Adds pause/unpause functionality for Sample.

Closes #329.

### DIFF
--- a/Sample.h
+++ b/Sample.h
@@ -95,12 +95,15 @@ public:
 	}
 
 
-	/** Resets the phase (the playhead) to the start position, which will be 0 unless set to another value with setStart();
-	*/
+	/** 
+	 * Resets the phase (the playhead) to the start position, which will be 0 unless set to another value with setStart();
+	 * Also unpauses playback if it's paused.
+	 */
 	inline
 	void start()
 	{
 		phase_fractional = startpos_fractional;
+		if(paused) setPaused(false);
 	}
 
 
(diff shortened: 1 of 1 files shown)


positron96:
I haven't tested it yet, so a draft for now.

tomcombriat:
Hi, 
Looks good at a first glance, I guess we should wait for your testing before a more thoughtful review ?

positron96:
ok, seems to work in simple tests

Sample.h
```diff
@@ -114,6 +117,20 @@ class Sample
...
+	 */
+	inline
+	void setPaused(bool to_pause = true) 
+	{
+		paused = to_pause; 
+	}
+
+	/** Checks whether playback is paused. */
+	inline
+	bool isPaused() { return paused; }
```
tomcombriat:
```suggestion
	bool isPaused() const { return paused; }
```

tomcombriat: I suggest adding `const` here.

tomcombriat (commented):
Looks good to me, thanks!

Just added a small optimization comment, otherwise, except if the others have some comments, will merge in a few day.

Sample.h
positron96: thanks!

positron96: Huh, weird, I thought the change already got into the codebase so didn't change it myself.

tomcombriat: I suggested it but did not apply it I case you would disagree, I applied it now. Waiting for the tests to complete and will merge. Thanks!

tomcombriat: (I also merged another PR with this one, which also dealt with the `Sample` class, explaining the second commit.)

